PR #1, opened 2018-07-18
US98453 Add d2l-enrollment-card

ryantmer:
This is effectively just the existing `d2l-course-image-tile`, but converted over to be `d2l-card`-based, and having the style updated to look as correct as possible - not including the new information that will soon be added to the card.


components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -0,0 +1,826 @@
...
+						this._image = imageEntity;
+					}
+				}
+				if (organization.hasSubEntityByRel(this.HypermediaRels.organizationHomepage)) {
+					var homepageEntity = organization.getSubEntityByRel(this.HypermediaRels.organizationHomepage);
+					this._organizationHomepageUrl = homepageEntity
+						&& homepageEntity.properties
+						&& homepageEntity.properties.path;
+				} else {
+					// If the user doesn't have access, don't animate image/show menu/underline on hover
```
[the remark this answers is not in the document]
ryantmer: Ah, yes - it's for the image saturation/zoom effect. After some discussion with Dave B, we're not totally sure if that's going to be a thing that the `d2l-card` takes care of or not - so I'll re-add here in the meantime, until we've figured that one out. Thanks!

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -0,0 +1,826 @@
...
+					<d2l-offscreen>[[_overlayAnnounceText]]</d2l-offscreen>
+				</div>
+				<div hidden$="[[!_showUpdateCount]]">
+					<d2l-offscreen>[[localize('courseTile.updates')]]</d2l-offscreen>
+					<span class="update-text-box">[[_updateString]]</span>
+				</div>
+			</div>
+		</d2l-card>
+
+		<div class="flex menu-area">
```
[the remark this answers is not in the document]
ryantmer: Twas indeed! Yeah, and I think it's something that should have always been the case. From a keyboard-navigation perspective, what we really have is [link to the course], [button to open menu], [button to unpin (if applicable)]. Previously, that was not the order things were in the DOM, but now it's more semantically correct - and the buttons are moved to the top-right of the image with some positioning instead.

ryantmer:
Note that this will require https://github.com/BrightspaceUI/tile/pull/60 before the tests pass, due to a mistake I made over in the `d2l-card` implementation.

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -0,0 +1,874 @@
...
+						}
+
+						var newSrcSet = this.getImageSrcset(e.detail.image, 'tile');
+						if (newSrcSet) {
+							imagePreloader.setAttribute('srcset', newSrcSet);
+						}
+
+						imagePreloader.setAttribute('sizes', this.$$('d2l-course-image').getTileSizes());
+						break;
+					case 'success':
```
m6jones: The tests assume that this event will show an overlay by setting `this._hasOverlay = true` then in 1001 seconds the test assumes this overlay will be hidden. Test fails on the hiding the overlay.

This can be solved in `_displaySetImageResult(true)` however, I don't see when this event shows the overlay so why should it hide it?

ryantmer: This is the stuff that happens after you change the image on the tile - not sure what's up with the failing tests, although I did slightly change the DOM structure so it could be that. Looking into it!

The overlay is being shown in `refreshImage`, which I believe is being called by the consumer (although I can't entirely remember to be honest). This is something that I forgot to check is still working correctly (setting images, that is), so I'm glad the tests reminded me!

ryantmer: Turned out to be an easy fix - I'd added another `div` to the DOM, so the `nth-of-type` selectors just had to be updated :)

ryantmer:
The plan here - once this is merged, I've got a PR that will move My Courses over to use this component instead of the `1.x-hybrid-components/d2l-course-image-tile`. That will be released (probably Monday) via BSI. After that, @m6jones's changes to make this entity-driven can be incorporated as well. Following that, it will be good to break this up a bit, and add the new information needed for learner dashboard - but the main advantage, this is no longer within My Courses itself, yay! Less risk and all that.

Also going to add a demo and docs for this component once it's broken down a bit (US98455).

ryantmer:
Urgh, also the test failure there is Safari on Sauce, which is currently being flaky and timing out (just doesn't run any tests at all). I've run the tests manually via Sauce, and all is well.

PR #2, opened 2018-07-20
Majones/us98454/covert d2l enrollment card to entity driven component

m6jones:


components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -79,17 +79,6 @@
...
-			.course-code-text,
-			.separator-icon,
-			.semester-text {
-				display: none;
-			}
-			:host([show-course-code]) .course-code-text,
-			:host([show-course-code]) .separator-icon,
-			:host([show-semester]) .semester-text {
-				display: inline-block;
-			}
```
m6jones: I removed this type of hiding since the host no longer has those attributes. Is this more preferable over using hidden&="*"?

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -287,18 +276,16 @@
...
 					[[_organization.properties.name]]
 					<div>
-						<span class="course-code-text uppercase">
-							[[_organization.properties.code]]
-							<d2l-icon hidden$="[[!_showSeparator]]" class="separator-icon" icon="d2l-tier1:bullet"></d2l-icon>
-						</span>
-						<span class="semester-text">[[_semesterName]]</span>
+						<span hidden$="[[!_presentation.properties.ShowCourseCode]]" class="course-code-text uppercase">[[_organization.properties.code]]</span>
+						<d2l-icon hidden$="[[!_showSeparator]]" class="separator-icon" icon="d2l-tier1:bullet"></d2l-icon>
+						<span hidden$="[[!_presentation.properties.ShowSemester]]" class="semester-text">[[_semesterName]]</span>
```
m6jones: I had issues with this updating properly since it is using a sub property. Is it better to take this out into it's own property? I figured that we will break this part into it's own component where we will pass either the presentationHref or the presentation object.

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -352,44 +339,50 @@
...
-					type: Boolean,
-					value: false,
-					reflectToAttribute: true
-				},
-				showSemester: {
+				_pinned: {
 					type: Boolean,
 					value: false,
-					reflectToAttribute: true
+					observer: '_handlePinnedChange'
```
m6jones: _handlePinnedChange basically adds a attribute to :host called pinned depending if it is true or false.

I did it this way since d2l-my-courses formats on the pinned attribute using a css selector so passing the information through the attribute seemed useful. However, I didn't want parent components being able to manually pin courses by adding an attribute so doing it this way makes that attribute read only.

This way worked better than having a property pinned that was readonly and reflected to attribute. But I feel like this should be the way to do it.

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -758,10 +802,10 @@
 				});
 			},
 			_onFocus: function() {
-				this.focused = true;
+				this.setAttribute('focused', '');
```
m6jones: I took away the focused property and added this to follow the pinned attribute pattern.

test/d2l-enrollment-card/d2l-enrollment-card.js
```diff
@@ -306,22 +335,35 @@ describe('d2l-enrollment-card', () => {
 		beforeEach(done => loadEnrollment(done));
 
 		it('should show the semester if the showSemester is set', () => {
-			component.showSemester = true;
+			presentationEntity.properties.ShowSemester = true;
+			component._presentation = {};
```
m6jones: Requiring this line to force a refresh. This to me is a red flag that component can do something better. Any guidance would be great. Thank you

wct.conf.json
```diff
@@ -14,17 +14,17 @@
...
           "browserName": "firefox",
-          "platform": "OS X 10.12",
+          "platform": "OS X 10.13",
           "version": ""
         },
         {
           "browserName": "safari",
-          "platform": "OS X 10.12",
+          "platform": "OS X 10.13",
           "version": ""
```
m6jones: Comment: https://d2l.slack.com/archives/C0PHG3QB0/p1532353616000478

components/d2l-enrollment-card/d2l-enrollment-card.html
ryantmer: Yeah, I figured `readOnly` + `reflectToAttribute` would have been the way to go, but if that wasn't reliable then this makes sense to me as well :+1:

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -352,44 +339,50 @@
...
-				},
-				showSemester: {
+				_pinned: {
 					type: Boolean,
 					value: false,
-					reflectToAttribute: true
+					observer: '_handlePinnedChange'
 				},
-				startedInactive: {
+				_startedInactive: {
```
ryantmer: This one is [also used by My Courses](https://github.com/Brightspace/d2l-my-courses-ui/blob/master/src/d2l-my-courses-content/d2l-my-courses-content-behavior.html#L390) to do some formatting, so I think will need a similar `observer` or readOnly/reflectToAttribute that `_pinned` has.

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -352,44 +339,50 @@
...
 					type: Boolean,
-					value: false,
-					reflectToAttribute: true
+					value: false
 				},
-				tileSizes: {
+				_tileSizes: {
 					type: Object,
-					value: function() { return {}; }
+					value: function() {
```
ryantmer: Legit - I think ostensibly, this is _supposed_ to do more than it currently did anyway, but that can be an investigation for another time. In the meantime, I'm glad this worked out 👌

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -588,22 +582,44 @@
...
 					&& semesterName.length > 0;
 			},
-			_computeShowUpdateCount: function(courseUpdatesConfig, updateCount) {
-				return !!courseUpdatesConfig && updateCount > 0;
+			_computeShowUpdateCount: function(presentation, updateCount) {
+				var courseUpdateConfig = presentation
+					&& presentation.properties
+					&& (presentation.properties.ShowUnattemptedQuizzes
+					|| presentation.properties.ShowDropboxUnreadFeedback
+					|| presentation.properties.SowUngradedQuizAttempts
```
ryantmer: Typo here - missing the "h" in `ShowUngradedQuizAttempts`

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -588,22 +582,44 @@
...
+					|| presentation.properties.ShowUnreadDiscussionMessages
+					|| presentation.properties.ShowUnreadDropboxSubmissions);
+
+				return courseUpdateConfig && updateCount > 0;
 			},
 			_computeUpdateString: function(updateCount) {
 				return updateCount > 99 ? '99+' : updateCount;
 			},
+			_computeShowSemester: function(presentation) {
+				return !!presentation
```
ryantmer: Curious about the `!!` here - is there a circumstance where `presentation` was a non-object?

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -588,22 +582,44 @@
...
+					&& !!presentation.properties
+					&& !!presentation.properties.ShowSemester;
+			},
+			_handlePinnedChange: function(pinned) {
+				if (pinned) {
+					this.setAttribute('pinned', '');
+				} else {
+					this.removeAttribute('pinned', '');
```
ryantmer: I believe `removeAttribute` doesn't need the second parameter, just the name of attribute to remove.

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -628,17 +644,41 @@
...
+					return Promise.resolve();
+				}
+
+				return this._fetchSirenEntity(presentationHref)
+					.then(function(presentation) {
+						this._presentation = presentation;
+					}.bind(this))
+					.then(function() {
+						return Promise.all([
+							this._fetchSemester(),
```
ryantmer: Hm, at this point we have the `_fetchSemester` and `_fetchNotifications` calls both here, and in _handleEnrollmentData. This makes sense, of course, since a change to either `href` or `presentationHref` could mean we need to recalculate these things, but I'm wondering if it makes sense to instead make these two functions first-class `observers` themselves? Might be beyond the scope of this change though, as worst-case we just end up fetching something twice (which will be deduped and cached anyway, so really only fetching it once).

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -663,19 +707,19 @@
...
-						if (this.courseUpdatesConfig.showUnattemptedQuizzes) {
+						if (this._presentation.properties.ShowUnattemptedQuizzes) {
 							total += notifications.properties.UnattemptedQuizzes;
 						}
-						if (this.courseUpdatesConfig.showDropboxUnreadFeedback) {
+						if (this._presentation.properties.ShowDropboxUnreadFeedback) {
 							total += notifications.properties.UnreadAssignmentFeedback;
 						}
-						if (this.courseUpdatesConfig.showUngradedQuizAttempts) {
+						if (this._presentation.properties.SowUngradedQuizAttempts) {
```
ryantmer: Missing the "h" here as well

components/d2l-enrollment-card/d2l-enrollment-card.html
ryantmer: I'm actually hoping we won't even need this much longer, although I couldn't get `:focus` to work correctly on the host element :(

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -758,10 +802,10 @@
 				});
 			},
 			_onFocus: function() {
-				this.focused = true;
+				this.setAttribute('focused', '');
 			},
 			_onBlur: function() {
-				this.focused = false;
+				this.removeAttribute('focused', '');
```
ryantmer: (same `removeAttribute` comment as above)

test/d2l-enrollment-card/d2l-enrollment-card.js
ryantmer: Hm - maybe `Polymer.dom.flush()`? It sometimes helps in situations like this, where a bound property doesn't update until after the test has finished running. This doesn't _always_ work, though - hence the large number of `setTimeout`s in these tests :sob: (granted, some of them could probably be replaced by `Polymer.dom.flush()` as well)

test/d2l-enrollment-card/d2l-enrollment-card.js
```diff
@@ -306,22 +335,35 @@ describe('d2l-enrollment-card', () => {
...
 			});
 		});
 
 		it('should not set the semester name if the show semester config is false', () => {
 			var spy = sandbox.spy(component, '_fetchSirenEntity');
+			presentationEntity.properties.ShowCourseCode = true;
+			component._presentation = presentationEntity;
 
-			component.showSemester = false;
+			component.dispatchEvent(new CustomEvent('d2l-enrollment-card-presentation-change', {
```
ryantmer: What's this event for? I'm not seeing a listener for it anywhere, but I could very well just not be seeing it 👓

components/d2l-enrollment-card/d2l-enrollment-card.html
m6jones: There was but I believe that is no longer possible as it should be.

components/d2l-enrollment-card/d2l-enrollment-card.html
m6jones: That is a good idea. They should be first class observers and even the http.request cache will stop the multiple calls.

test/d2l-enrollment-card/d2l-enrollment-card.js
m6jones: Whoops, that was put in when I was trying to find a solution. I found a far better one.

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -687,7 +731,7 @@
 					});
 			},
 			_fetchSemester: function() {
-				if (!this._semesterUrl || !this.showSemester) {
+				if (!this._semesterUrl || !this._presentation.properties.ShowSemester) {
```
njostonehouse: We might want to grab the Semester name even if we're not going to show it to the user. I think if a consumer calls `_fetchSemester` they would expect the value to get set.

ryantmer: Entirely orthogonal to the point you're making, but consumers shouldn't be calling `_fetchSemester` 😉

That said, we can avoid the network call (which will likely be in that initial-loading-everything-is-fighting-for-resources stage) this way - not a big benefit, but I figure if `!showSemester`, then there hopefully isn't a need to have the name anyway?

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -588,22 +583,51 @@
...
+					|| presentation.properties.ShowUngradedQuizAttempts
+					|| presentation.properties.ShowUnreadDiscussionMessages
+					|| presentation.properties.ShowUnreadDropboxSubmissions);
+
+				return courseUpdateConfig && updateCount > 0;
 			},
 			_computeUpdateString: function(updateCount) {
 				return updateCount > 99 ? '99+' : updateCount;
 			},
+			_computeShowSemester: function(presentation) {
```
ryantmer: Missed this earlier - this appears to be unused

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -287,18 +276,16 @@
...
 			<div class="flex">
 				<div class="course-text">
 					[[_organization.properties.name]]
 					<div>
-						<span class="course-code-text uppercase">
-							[[_organization.properties.code]]
-							<d2l-icon hidden$="[[!_showSeparator]]" class="separator-icon" icon="d2l-tier1:bullet"></d2l-icon>
-						</span>
-						<span class="semester-text">[[_semesterName]]</span>
+						<span hidden$="[[!_presentation.ShowCourseCode]]" class="course-code-text uppercase">[[_organization.properties.code]]</span>
```
ryantmer: Huh - this still looks correct in all the different cases right? I'm not sure why it was so weirdly laid out before, with the the code and icon within one span, and the semester name in another...

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -588,22 +585,44 @@
...
+			_handlePinnedChange: function(pinned) {
+				if (pinned) {
+					this.setAttribute('pinned', '');
+				} else {
+					this.removeAttribute('pinned');
+				}
+			},
+			_handleStartedInactiveChange: function(startedInactive) {
+				if (startedInactive) {
+					this.setAttribute('startedInactive', '');
```
ryantmer: The selector in My Courses [is looking for](https://github.com/Brightspace/d2l-my-courses-ui/blob/39d59c55a6ab8a5441445f5e4ffdd7be0b1feac6/src/d2l-my-courses-content/d2l-my-courses-content-behavior.html#L390) `started-inactive` here with the dash, rather than `startedInactive`, without the dash.

components/d2l-enrollment-card/d2l-enrollment-card.html
m6jones: To me it looks correct and doesn't change anything. I didn't try all possible cases though.

ryantmer: Yeah, I'm not too concerned :) Sounds good.

PR #5, opened 2018-07-26
Minor style fixes

ryantmer:
- Update count should be top-right, not centre-right
- Menu should be 12px away from top-right corner, not 15px
- Make cards full-height (not 100% correct yet)

Cards in a single row should all be the same height, i.e. the height of the tallest card. This fixes doesn't make it perfect, but it's much closer. Making it perfect may require a change in `d2l-card` itself.


ryantmer:
The new GitHub Travis integration not picking up successes on re-runs is started to grind my gears 😰

PR #11, opened 2018-08-10
Majones/us97551/due date display on card

m6jones:
This change adds due dates to cards as well as the overdue badge if a course is over due.
It also add course completion date and the completion badge.


components/d2l-user-activity-usage/d2l-user-activity-usage.html
```diff
@@ -0,0 +1,142 @@
...
+
+		</style>
+		[[_dateText]]
+	</template>
+	<script>
+		Polymer({
+			is: 'd2l-user-activity-usage',
+
+			properties: {
+				userActivityUsageHref: {
```
ryantmer: Question here - a) should this just be `href`, and b) should this component exist in a different repo? Generally, the BHC components have been organized on a per-repo basis on what type of entity drives them. Should this be in an `activities` repo or something? (We can also defer this decision too, as I'm not totally clear at this point whether that makes sense for this use case.)

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -14,8 +14,11 @@
 <link rel="import" href="../../../d2l-organization-hm-behavior/d2l-organization-hm-behavior.html">
 <link rel="import" href="../../../d2l-organizations/d2l-organization-info.html">
 <link rel="import" href="../../../d2l-tile/d2l-card.html">
+<link rel="import" href="../../../d2l-status-indicator/d2l-status-indicator.html">
 <link rel="import" href="../../../d2l-organizations/d2l-organization-updates.html">
 <link rel="import" href="../../../siren-parser-import/siren-parser.html">
+<link rel="import" href="../../../siren-parser-import/siren-parser.html">
```
ryantmer: Duplicate import here :)

components/d2l-user-activity-usage/localize-behavior.html
```diff
@@ -0,0 +1,129 @@
...
+						},
+						'ko': {
+							'brackets': '({content})',
+							'dueOn': '닫힘 일시: {dateTime}',
+							'completedOn': '열림 일시 {dateTime}',
+							'dueTomorrow': '비활성',
+						},
+						'nb': {
+							'brackets': '({content})',
+							'dueOn': 'Closed {dateTime}',
```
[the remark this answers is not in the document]
m6jones: Right, I wanted to talk to about where I should look for those being translated.

components/d2l-user-activity-usage/d2l-user-activity-usage.html
m6jones: I've been thinking about that too. I think putting it into the activities repo should happen. Originally, this only dealt with due dates and was called d2l-enrollment-due-date. However, I think it makes sense to include, completion date, start date, end date and so on. So that is why I changed the name. Changing the repo also make sense I think especially when the other functionalities join.

test/d2l-user-activity-usage/d2l-user-activity-usage.js
```diff
@@ -0,0 +1,270 @@
...
+					],
+					properties: {
+						date: '2017-08-01T04:00:00.000Z'
+					}
+				}
+			];
+
+			loadUserActivityUsage(entities);
+
+			setTimeout(() => {
```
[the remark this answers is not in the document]
m6jones: Flushing the dom didn't work.

PR #12, opened 2018-08-14
Majones/accessibility and d2l-card integration

m6jones:
Requires https://github.com/BrightspaceHypermediaComponents/enrollments/pull/11


components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -718,6 +792,26 @@
...
+					accessibility.organizationCode,
+					accessibility.semesterName,
+					accessibility.badge,
+					accessibility.userActivityUsageInfo
+				];
+				return textData.filter(function(text) {
+					return text && typeof text === 'string';
+				}).join(', ');
+			},
+			_accessibilityDataReset: function() {
```
m6jones: I cannot figure out a way where this can be removed.

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -718,6 +792,26 @@
...
+			_accessibilityDataToString: function(accessibility) {
+				if (!accessibility || !accessibility.organizationName) {
+					return this.localize('courseClosed');
+				}
+				var textData = [
+					accessibility.organizationName,
+					accessibility.organizationCode,
+					accessibility.semesterName,
+					accessibility.badge,
+					accessibility.userActivityUsageInfo
```
m6jones: These will populate the text field to give the same information as the visual user. Is it too much?

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -217,55 +210,96 @@
...
+
+				<button
+					hidden$="[[!_pinned]]"
+					class="pin-indicator"
+					on-tap="_pinClickHandler"
+					on-keypress="_pinPressHandler"
+					aria-label$="[[_pinButtonLabel]]">
+					<d2l-icon icon="d2l-tier1:pin-filled"></d2l-icon>
+				</button>
+			</div>>
```
dbatiste: Looks like an extra `>` on this closing tag.

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -242,45 +224,82 @@
 				href="[[_notificationsUrl]]"
 				presentation-href="[[presentationHref]]">
 			</d2l-organization-updates>
+			<div slot="actions" class="menu-area">
```
ryantmer: Kind of confused by this change actually - what was the issue with the menu being the sibling to the card in the DOM? The intent there was that screen readers (well, NVDA was the only one I tested) could follow the natural DOM order, announcing each item in sequence.

m6jones: Dave can answer better.
From my understanding, it hasn't changed order much as I believe the card it self got an extra layer. So it is still a sibling to the "base card" - the container holding the `header` and the `content`.

The reason behind the change is so that menus on all the cards are consistent and use the same styling which is provided by the d2l-card. This part is still being worked on but when it is finished I believe the class`menu-area` will no longer be required.

bower.json
```diff
@@ -29,7 +29,7 @@
     "d2l-offscreen": "^3.0.3",
     "d2l-organization-hm-behavior": "Brightspace/organization-hm-behavior#^2.0.1",
     "d2l-organizations": "BrightspaceHypermediaComponents/organizations#0.0.3-hybrid",
-    "d2l-tile": "^4.1.0",
+    "d2l-tile": "^4.1.1",
```
dbatiste: Should bump to `4.1.2` to pick up latest resize-observer reference that will not conflict with requireJS.

components/d2l-enrollment-card/d2l-enrollment-card.html
dbatiste: Yep.  Also, the actions slot will be placed on-top (z-index) of the link area as well.  Otherwise the link area will intercept the click and the user will not be able to interact with the button or menu opener.

components/d2l-enrollment-card/d2l-enrollment-card.html
m6jones: The reason is different properties should be in the string at different positions. Since they are added at random times to the card. I can't think of a way to make just a string work.

components/d2l-enrollment-card/localize-behavior.html
```diff
@@ -27,8 +27,8 @@
 							'coursePinButton': '{course} is pinned. Unpin course',
 							'courseSettings': '{course} course settings',
 							'courseTile.updates': 'updates:',
-							'overlay.courseClosed': 'Closed',
-							'overlay.courseClosedOn': 'Closed {dateTime}',
+							'courseClosed': 'Closed',
+							'courseClosedOn': 'Closed {dateTime}',
 							'overlay.courseOpens': 'Opens On',
```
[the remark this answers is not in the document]
m6jones: I just saw that file! I can remove them all.

demo/data/organizations.json
```diff
@@ -24,7 +24,9 @@
         "rel":[
           "https://api.brightspace.com/rels/organization-homepage"
         ],
-        "href":"http://example.com"
+        "properties": {
```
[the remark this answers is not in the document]
m6jones: I'll look into this.

components/d2l-enrollment-card/d2l-enrollment-card.html
ryantmer: Ahhh, yes, I see what you mean. Hmm. String templating might work? But also might be hella complicated and not worth it!

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -209,7 +189,9 @@
 			}
 		</style>
 
-		<d2l-card on-tap="_onTapCard" on-keypress="_onKeypressCard" tabindex="0">
+		<d2l-card subtle
+			href="[[_organizationHomepageUrl]]"
```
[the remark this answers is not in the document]
m6jones: It does still work. The anchor goes away but the rest is still visible.

demo/data/organizations.json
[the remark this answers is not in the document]
m6jones: It is path because it does a relative path.

test/d2l-enrollment-card/d2l-enrollment-card.js
```diff
@@ -465,4 +465,86 @@ describe('d2l-enrollment-card', () => {
...
+			setTimeout(() => {
+				expect(component._accessibilityData.badge).to.equal('overdue');
+				var cardText = component.$$('d2l-card').getAttribute('text');
+				expect(cardText).to.contain('overdue');
+				done();
+			});
+
+		});
+
+		it('No organization name', () => {
```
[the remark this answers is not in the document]
m6jones: Currently, if we don't get organization name things won't break. Should I trigger an error? Right now, I think it will just display the card with no name.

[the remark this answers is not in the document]
ryantmer: Nah, tis quite fine - just saying that this test is testing a scenario that can't be hit in a normal way!

demo/data/organizations.json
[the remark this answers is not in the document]
ryantmer: Ah! Apparently there is both an `organization-homepage` sub-entity, and an `organization-homepage` link... we do indeed use the sub-entity in `d2l-course-tile` though, you're totally right!

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -242,45 +224,82 @@
...
+
+				<button
+					hidden$="[[!_pinned]]"
+					class="pin-indicator"
+					on-tap="_pinClickHandler"
+					on-keypress="_pinPressHandler"
+					aria-label$="[[_pinButtonLabel]]">
+					<d2l-icon icon="d2l-tier1:pin-filled"></d2l-icon>
+				</button>
+			</div>
```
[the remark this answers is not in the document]
m6jones: This part has been fixed since it was below this point that was the problem.

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -76,35 +77,16 @@
...
-				margin-right: auto;
+				margin-right: 5px;
 			}
-			.menu-area > d2l-dropdown-more {
+			:host .menu-area > d2l-dropdown-more {
 				opacity: 0;
-				margin-top: 0;
-				transition: opacity 0.5s, margin-top 0.5s;
 			}
 			:host:hover .menu-area > d2l-dropdown-more,
```
dbatiste: Apologies for chiming in late here.  FYI: once I have my actions changes done (imminent), I think you will be able to remove this custom styling, and there may also be some JS that can be removed since the behavior of showing the dropdown opener will be handled automatically by adding the `visible-on-ancestor` attribute to the opener.

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -209,7 +191,10 @@
 			}
 		</style>
 
-		<d2l-card on-tap="_onTapCard" on-keypress="_onKeypressCard" tabindex="0">
+		<d2l-card
+			subtle
+			href="[[_organizationHomepageUrl]]"
+			text="[[_accessibilityDataToString(_accessibilityData)]]">
 			<div slot="header" aria-hidden="true">
```
dbatiste: I am curious about the `aria-hidden` here.  What was being announced previously that we did not want?  Now that the dropdown menu is in the correct slot, none of that should be announced.  Anyway, I am curious so I can inform other consumers if it's something they will need to do, or if there is some change needed in `d2l-card`.

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -217,55 +210,96 @@
...
+				</d2l-organization-info>
+				<d2l-user-activity-usage href="[[_userActivityUsageUrl]]"></d2l-user-activity-usage>
+			</div>
 			<d2l-organization-updates
 				slot="footer"
 				href="[[_notificationsUrl]]"
 				presentation-href="[[presentationHref]]">
 			</d2l-organization-updates>
+			<div slot="actions" class="menu-area">
+				<d2l-dropdown-more label="[[_courseSettingsLabel]]">
```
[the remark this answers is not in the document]
dbatiste: A brief follow-up on this... we'll want to switch this to the following very shortly:
```
<d2l-dropdown-more label="[[_courseSettingsLabel]]" translucent visible-on-ancestor>
```
The `translucent` option has been added to `d2l-button-icon`, `d2l-dropdown-more`, and `d2l-dropdown-context-menu`.  The `VisibleOnAncestorBehavior` has been added to them as well, and the card container defines a `d2l-visible-on-ancestor-target`, so to enable that behavior on button/dropdown basis in the card actions simply requires adding the `visible-on-ancestor` attribute as above on the desired buttons/dropdown openers.

components/d2l-enrollment-card/d2l-enrollment-card.html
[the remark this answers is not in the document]
dbatiste: We will want to change this from `button` element to `d2l-button-icon`.  This should enable a bunch of cleanup of styles and possibly some JS.

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -642,17 +637,6 @@
 			_onBlur: function() {
 				this.removeAttribute('focused');
```
dbatiste: Looks like this is being used for the custom visibility behavior that we can get rid of as mentioned above.  If I get some time, I can probably help with some of this cleanup.

components/d2l-enrollment-card/d2l-enrollment-card.html
m6jones: I believe this has been carried over by the previous iteration of the couse tile. This should be tested without to see if it is still required. If it is I think the `d2l-card` should handle that?

components/d2l-enrollment-card/d2l-enrollment-card.html
m6jones: Awesome, I do plan on going through and cleaning up some of the unnecessary code with all the new features of `d2l-card` I'll make sure you are one of the reviewers of that.

dbatiste: Yeah.  I am going through this process for image-banner-overlay currently (which is/was doing the same thing).

PR #13, opened 2018-08-17
Combine updates and moves everything left

m6jones:
I noticed an issue with course code and semester moving to the same line as the title. Fix for that is https://github.com/BrightspaceHypermediaComponents/organizations/pull/11

There is another issue where action drop down menu loses its z-index when focused. This PR removes a lot of the custom CSS to make it easier for the d2l-card to fix this issue.


components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -181,14 +181,19 @@
 				overflow-wrap: break-word; /* replaces 'word-wrap' in Firefox, Chrome, Safari */
 				display: flex;
 				flex-direction: column;
-				width: 100%;
-				text-align: center;
+				margin: -0.35rem 0 -0.1rem -0.05rem;
```
m6jones: This padding is temp solution to clean the cards white space. Dave B is working on a far better solution.

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -181,14 +181,19 @@
...
 			}
 			.badge {
 				background-color: #FFFFFF;
 				box-shadow: 0 0 0 2px #FFFFFF;
 				padding: 0.2rem 0.5rem 0.15rem 0.5rem;
 			}
+
+			d2l-organization-updates {
+				float: right;
+				margin: -0.7rem 0 0 0.2rem;
```
m6jones: This lines up the bottom of the text to the title.

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -60,34 +61,9 @@
...
-				width: 35px;
-				height: 35px;
-			}
-			.pin-indicator:hover,
-			.pin-indicator:focus {
-				color: rgba(255,255,255,1);
-				background: var(--d2l-color-celestine);
-			}
-
 			.menu-area > * {
```
dbatiste: I suspect this may not be needed.  Will need to check.  The card has this:
```
.d2l-card-actions ::slotted(*) {
		margin-left: 0.3rem;
}
```

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -224,13 +212,9 @@
 					<d2l-user-activity-usage href="[[_userActivityUsageUrl]]"></d2l-user-activity-usage>
 				</d2l-card-content-meta>
 			</div>
-			<d2l-organization-updates
-				slot="footer"
-				href="[[_notificationsUrl]]"
-				presentation-href="[[presentationHref]]">
-			</d2l-organization-updates>
+
 			<div slot="actions" class="menu-area">
```
dbatiste: I think we should be able to get rid of the `class="menu-area"`.

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -257,14 +241,15 @@
 					</d2l-dropdown-menu>
 				</d2l-dropdown-more>
 
-				<button
+				<d2l-button-icon
+					slot="actions"
```
dbatiste: It looks like this `slot="actions"` should not be here since the slot is already specified for the parent `div`.

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -215,6 +196,13 @@
 
 			<d2l-status-indicator class="badge" slot="badge" state="[[_badgeState]]" text="[[_badgeText]]" hidden$="[[!_badgeText]]"></d2l-status-indicator>
 
+			<d2l-organization-updates
```
dbatiste: Not sure if it makes any difference to group this and the next div and only specify `slot="content"` once?? I'm not really sure if it makes a difference one way or another.

dbatiste:
Looks like we can get rid of those focus and blur handlers now too!

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -60,34 +61,9 @@
...
-			.pin-indicator:hover,
-			.pin-indicator:focus {
-				color: rgba(255,255,255,1);
-				background: var(--d2l-color-celestine);
-			}
-
 			.menu-area > * {
 				margin-right: 5px;
 			}
-			:host .menu-area > d2l-dropdown-more {
```
dbatiste: Ya!

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -60,35 +61,6 @@
 				display: flex;
 			}
 
-			.pin-indicator {
```
ryantmer: 🙌

Loving seeing all this go

components/d2l-enrollment-card/d2l-enrollment-card.html
ryantmer: Not sure what you mean here - is the idea to squish the `-updates` component closer to the `-meta` text?

(As an aside, I personally prefer the non-short-form `margin-top/margin-blah`, as I always forget what the order is... but maybe I should just remember the order and stop complaining 😝)

components/d2l-enrollment-card/d2l-enrollment-card.html
ryantmer: Huh, this is news to me! I figured the second `slot="content"` would end up overriding the first one. Neat!

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -224,48 +208,43 @@
...
+							on-d2l-menu-item-select="_pinClickHandler"
+							hidden$="[[!_pinned]]"
+							text="[[localize('unpin')]]">
+						</d2l-menu-item>
+					</d2l-menu>
+				</d2l-dropdown-menu>
+			</d2l-dropdown-more>
+
+			<d2l-button-icon
+				slot="actions"
```
ryantmer: Oh dang, yeah I definitely prefer this multi-`slot` approach here. Allows an arbitrary number of actions to be added, which is great!

demo/data/organizations/2.json
```diff
@@ -0,0 +1,52 @@
+{
+  "class":[
+     "active",
+     "course-offering"
+  ],
+  "properties":{
+     "name":"Rindfleischetikettierungsueberwachungsaufgabenuebertragungsgesetz",
```
ryantmer: Google translates this as "Beef meat labeling surveillance task transfer law", is this a real word? 🤣

components/d2l-enrollment-card/d2l-enrollment-card.html
m6jones: Think of N E S W compass.
Basically, it lines the notification text with the title text. Bob wanted the bottom of both text to line up.

demo/data/organizations/2.json
m6jones: I googled longest german word :D

components/d2l-enrollment-card/d2l-enrollment-card.html
ryantmer: Hah, the compass thing is a great idea actually.

And I see! Sounds good, just wasn't able to picture it. Thanks!

PR #15, opened 2018-08-24
Revamped the polymer demos

m6jones:
Reorganized and made a demo to illustrate the multiple forms an enrollment card can take.


demo/d2l-enrollment-card/d2l-enrollment-card-demo.html
```diff
@@ -20,76 +20,183 @@
...
 				<template>
-					<d2l-enrollment-card href="../data/enrollments/2.json" presentation-href="../data/presentation.json">
+					<d2l-enrollment-card href="../data/enrollments/inactive/1.json" presentation-href="../data/base/presentation.json">
 					</d2l-enrollment-card>
 				</template>
 			</demo-snippet>
 
-			<h3>Completed Enrollment Card</h3>
-
+			<h3>Enrollment Card - Organization Info</h3>
```
ryantmer: This is possibly just a semantic nitpick, but would it make sense to have the full exercising of the `d2l-organization-info` component as demos in that repo instead? I realize it doesn't give the "full picture", granted, but it might be good to keep this demo a bit simpler.

ryantmer: Yeah, totally valid. My thought above is largely a thought exercise really - not something that needs to be changed, more just a consideration for further development; as components become reliant on other components, the fanout of possible demo options becomes large, so worth keeping in mind :)

PR #19, opened 2018-08-30
New Course Highlight

m6jones:
Adds a blue indicator at the top of the card to show that this enrollment has not been accessed.
image

### DIFF
--- a/components/d2l-enrollment-card/d2l-enrollment-card.html
+++ b/components/d2l-enrollment-card/d2l-enrollment-card.html
@@ -134,14 +134,14 @@ Polymer-based web component for a course/enrollment card.
 			}
 
 			.alert-colour-circle {
-				height: 20px;
-				width: 20px;
-				border-radius: 20px;
-				border: 3px solid white;
-				background-color: #ffce51;
+				height: 1rem;
+				width: 1rem;
+				border-radius: 50%;
+				border: 2px solid white;
+				background-color: var(--alert-circle-colour);
 				position: absolute;
-				top: -6px;
-				right: -6px;
+				top: -10px;
+				right: -10px;
 			}
 			[dir="rtl"] .alert-colour-circle {
 				right: auto;
@@ -186,8 +186,6 @@ Polymer-based web component for a course/enrollment card.
 						<d2l-icon></d2l-icon>
 					</div>
 				</div>
-
-				<div class="alert-colour-circle" hidden$="[[!_startedInactive]]"></div>
 			</div>
 
 			<d2l-status-indicator class="badge" slot="badge" state="[[_badgeState]]" text="[[_badgeText]]" hidden$="[[!_badgeText]]"></d2l-status-indicator>
@@ -247,7 +245,7 @@ Polymer-based web component for a course/enrollment card.
 			</d2l-button-icon>
 		</d2l-card>
 
-
+		<div class="alert-colour-circle" hidden$="[[!_showAlertCircle]]"></div>
 	</template>
 	<script>
 		Polymer({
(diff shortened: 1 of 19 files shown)


test/d2l-enrollment-card/d2l-enrollment-card.js
```diff
@@ -547,4 +547,22 @@ describe('d2l-enrollment-card', () => {
 
 	});
 
+	describe('New Enrollment Highlight', () => {
+
+		beforeEach(done => loadEnrollment(done));
+
+		afterEach(() => sandbox.reset());
+
+		it('See hightlight', done => {
```
ryantmer: tiny nit: typo

test/d2l-user-activity-usage/d2l-user-activity-usage.js
```diff
@@ -300,6 +300,34 @@ describe('d2l-user-activity-usage', () => {
...
+					rel: [
+						'https://api.brightspace.com/rels/date'
+					],
+					properties: {
+						date: '2017-08-01T04:00:00.000Z'
+					}
+				}
+			];
+
+			var eventSpy = sinon.spy();
```
ryantmer: `sandbox.spy` works too - not a big concern here since it's a local variable anyway, but I find it's a good practice to be in :) (I'm realizing, though, that there are other `sinon.spy`s in this file, so 🤷‍♀️)

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -437,6 +441,18 @@
 					&& organization.properties
 					&& this.localize('courseSettings', 'course', organization.properties.name);
 			},
+			_computeShowAlertCircle: function(startedInactive, newEnrollment) {
+				var colour = null;
+				if (startedInactive) {
+					colour = '#ffce51';
```
ryantmer: I'm thinking this is actually supposed to be [Citrine](http://design.d2l/style/colour/) - it probably was originally that colour, which has since been updated to be slightly different... but `#ffce51` and `#ffba59` are pretty similar.

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -247,7 +245,7 @@
 			</d2l-button-icon>
 		</d2l-card>
 
-
+		<div class="alert-colour-circle" hidden$="[[!_showAlertCircle]]"></div>
```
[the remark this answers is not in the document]
m6jones: So this is because the d2l-card would hide all overflow. So it would cut off if I didn't bring it out. However, this is no longer an issue with the card. Having it outside of the card actually made the indicator not follow the subtle movement. So I moved it back into the content slot.

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -134,14 +134,14 @@
...
+				height: 1rem;
+				width: 1rem;
+				border-radius: 50%;
+				border: 2px solid white;
+				background-color: var(--alert-circle-colour);
 				position: absolute;
-				top: -6px;
-				right: -6px;
+				top: -10px;
+				right: -10px;
```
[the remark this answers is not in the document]
dbatiste: Note that we'll need CSS selectors [like this](https://github.com/BrightspaceUI/tile/blob/master/d2l-card.html#L65) (to cover P1 / P2 + shady / shadow-dom scenarios).

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -437,6 +441,18 @@
 					&& organization.properties
 					&& this.localize('courseSettings', 'course', organization.properties.name);
 			},
+			_computeShowAlertCircle: function(startedInactive, newEnrollment) {
+				var colour = null;
+				if (startedInactive) {
```
dbatiste: It looks like the color is hard-coded here, in which case we can put these colors in CSS and select them based on a class or attribute selector, thereby avoiding the `updateStyles` call.  Yes, no?

m6jones: You are totally right! 🤦‍♀️

PR #20, opened 2018-09-04
US99502: Course cards show status badges and course metadata as per design and business logic definition

m6jones:
- Uses the new organization split.
- If no user activity then uses organizations end and start date.
- Insure that badge shows correctly based on business logic. (see story)
- Added inactive, completed, overdue, closed and disabled as read only boolean attributes. So css can sort on them.
- Changed started-inactive boolean attribute to disabled. Will still add started-inactive attribute to match the disabled state.
- Updated demos to show the new cases.
- Updated tests to test the new cases.

- [x] Requires: https://github.com/BrightspaceHypermediaComponents/organizations/pull/13


components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -134,14 +160,14 @@
...
 
 			.alert-colour-circle {
-				height: 1rem;
-				width: 1rem;
+				height: .75rem;
+				width: .75rem;
 				border-radius: 50%;
-				border: 2px solid white;
-				background-color: var(--alert-circle-colour);
+				border: 1px solid #f6f7f8;
```
m6jones: #f6f7f8 is not a named colour but currently the colour of the background.

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -134,14 +160,14 @@
...
+				background-color: var(--d2l-color-celestine);
 				position: absolute;
-				top: -10px;
-				right: -10px;
+				top: -0.375rem;
+				right: -0.375rem;
 			}
 			[dir="rtl"] .alert-colour-circle {
```
m6jones: - [ ] Requires the rtl for shady.

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -194,23 +223,28 @@
 				</div>
 			</div>
 
-			<d2l-status-indicator class="badge" slot="badge" state="[[_badgeState]]" text="[[_badgeText]]" hidden$="[[!_badgeText]]"></d2l-status-indicator>
+			<d2l-status-indicator aria-hidden="true" class="badge" slot="badge" state="[[_badgeState]]" text="[[localize(_badgeText)]]" hidden$="[[!_badgeText]]"></d2l-status-indicator>
```
m6jones: The anchor for the card has the accessibility attached to it so mute all other elements. This treats the card a a single element.

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -194,23 +223,28 @@
...
 
 			<d2l-organization-updates
 				combined
 				badge$="[[_badgeText]]"
 				slot="content"
 				href="[[_notificationsUrl]]"
-				presentation-href="[[presentationHref]]">
+				presentation-href="[[presentationHref]]"
+				aria-hidden="true">
```
m6jones: Same as above as the information for the voice reader is sent to the card.

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -267,25 +301,47 @@
...
+					value: false,
+					readOnly: true
+				},
+				closed: {
+					type: Boolean,
+					reflectToAttribute: true,
+					value: false,
+					readOnly: true
+				},
+				inactive: {
```
m6jones: These properties are read-only attributes so the user can style based on state. For example when filtering elements.

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -194,23 +223,28 @@
...
+				<d2l-organization-name href="[[_organizationUrl]]"></d2l-organization-name>
 				<d2l-card-content-meta>
-					<d2l-user-activity-usage href="[[_userActivityUsageUrl]]"></d2l-user-activity-usage>
+					<d2l-organization-info
+						class="block"
+						href="[[_organizationUrl]]"
+						presentation-href="[[presentationHref]]">
+					</d2l-organization-info>
+					<d2l-user-activity-usage class="block" href="[[_userActivityUsageUrl]]" override-to-default="[[_orgDateSlot]]">
+						<d2l-organization-date slot="default" href="[[_organizationUrl]]"></d2l-organization-date>
```
m6jones: Shady dom doesn't allow dynamic slot definitions. So I had to add `override-to-default` to get the same result.
If `d2l-user-activity-usage` has no information then `default` is shown. However, if `override-to-default` is true then always use default.

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -777,6 +828,25 @@
 				var accessiblity = this._accessibilityData;
 				this._accessibilityData = {};
 				this._accessibilityData = accessiblity;
+			},
+			_setBadgeText: function() {
```
m6jones: This was an observer for the used properties in this method. However, the shady dom couldn't handle it.

bower.json
```diff
@@ -29,8 +29,8 @@
     "d2l-menu": "^1.1.0",
     "d2l-offscreen": "^3.0.3",
     "d2l-organization-hm-behavior": "Brightspace/organization-hm-behavior#^2.0.1",
-    "d2l-organizations": "BrightspaceHypermediaComponents/organizations#0.0.6-hybrid",
-    "d2l-tile": "^4.1.4",
+    "d2l-organizations": "BrightspaceHypermediaComponents/organizations#0.1.0-hybrid",
+    "d2l-tile": "^4.1.7",
```
dbatiste: I've finished my task of renaming `d2l-tile` to `d2l-card`.  It is possible to list `d2l-card` as the dependency, although d2l-tile will continue to work.  It would be good to update though.

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -13,11 +13,13 @@
 <link rel="import" href="../../../d2l-offscreen/d2l-offscreen.html">
 <link rel="import" href="../../../d2l-organization-hm-behavior/d2l-organization-hm-behavior.html">
 <link rel="import" href="../../../d2l-organizations/d2l-organization-info.html">
+<link rel="import" href="../../../d2l-organizations/d2l-organization-updates.html">
+<link rel="import" href="../../../d2l-organizations/d2l-organization-date.html">
+<link rel="import" href="../../../d2l-organizations/d2l-organization-name.html">
 <link rel="import" href="../../../d2l-tile/d2l-card.html">
```
dbatiste: Same here - update these paths when switching to `d2l-card`.

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -43,14 +45,35 @@
...
+				Safari 6+,
+				Safari 6+ iOS,
+				Opera 15+
+				*/
+				-webkit-filter: grayscale(1);
+
+				/* Firefox 35+ */
 				filter: grayscale(1);
+
+				/* IE 6-9 */
```
dbatiste: Ha - only for IE6-9?  If so I say we ditch.

components/d2l-enrollment-card/d2l-enrollment-card.html
dbatiste: Are we missing a color from our palette?

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -60,6 +83,9 @@
 			.flex {
```
dbatiste: Is this no longer used?  

Aside: something to be careful of - if something did have this class (ex. `<div class="flex">`), if anything else on the page defined this style in traditional, non-scoped way (ex. LMS), those styles could affect this element when shady-DOM is enabled since the internal of the component would not really be isolated from the rest of the DOM.

components/d2l-enrollment-card/d2l-enrollment-card.html
m6jones: We are missing this one. How do we add a new one such that it is included in design.d2l? 
Or do I push back on the design team to use a proper colour?

dbatiste: I would think the palette would include a color for this, and usually when I see new colors I question it to be sure it's appropriate to do a "one-off".  What does the color mean in this case?

m6jones: It is the match the background colour of the my courses widget. So it isn't a "one-off".

dbatiste: I see, that's the tiles container.

dbatiste: Hmmm, `transparent` is not going to give the look we want.  Might need to leave as this for now, and perhaps follow up with someone about the color.

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -43,22 +45,41 @@
...
+
+				opacity: 0.5;
 			}
 			d2l-card[disabled]:hover,
 			d2l-card[disabled]:focus {
 				cursor: not-allowed;
 			}
 
+			d2l-card[disabled] d2l-organization-updates,
+			d2l-card[disabled] [slot=actions] {
```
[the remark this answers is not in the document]
m6jones: This is probably a specific use case. Looks really bad when they are visible in this case.

components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -176,6 +207,7 @@
 
 		<d2l-card
 			subtle
+			disabled$=[[disabled]]
```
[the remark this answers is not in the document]
m6jones: Whoops. I'll need to fix that.

PR #21, opened 2018-09-07
Sends an event when a course has a status of completed.

m6jones:


components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -538,6 +539,17 @@
 
 				this._accessibilityDataReset();
 			},
+			_handleCompletedChange: function(newCompleted, oldCompleted) {
+				if (!this._enrollment
+					|| !this._enrollment.hasLinkByRel
+					|| newCompleted === oldCompleted) {
```
ryantmer: Huh, does the `newCompleted === oldComplete` case actually happen? The observer should really only be firing if the value has changed 😮

m6jones: I feel like I read some where it does but from all my other experiences sort of tells me it doesn't. I'll test it to see.

PR #32, opened 2018-10-11
Resets the state if the enrollment changes.

m6jones:
Enrollment cards wouldn't reset their state. So if the sort order changed, the card would get a new `href` but all the badges would stay. So new course information but with old inactive badge

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -30,7 +30,7 @@
     "d2l-menu": "^1.1.0",
     "d2l-offscreen": "^3.0.3",
     "d2l-organization-hm-behavior": "Brightspace/organization-hm-behavior#^2.0.1",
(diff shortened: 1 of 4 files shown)


bower.json
```diff
@@ -30,7 +30,7 @@
     "d2l-menu": "^1.1.0",
     "d2l-offscreen": "^3.0.3",
     "d2l-organization-hm-behavior": "Brightspace/organization-hm-behavior#^2.0.1",
-    "d2l-organizations": "BrightspaceHypermediaComponents/organizations#^0.1.5-hybrid",
+    "d2l-organizations": "BrightspaceHypermediaComponents/organizations#^0.1.7-hybrid",
```
m6jones: Not related to this issue.

PR #34, opened 2018-10-15
Enrollment card is properly reset when new link is given.

m6jones:


components/d2l-user-activity-usage/d2l-user-activity-usage.html
```diff
@@ -100,6 +100,7 @@
 			},
 			_hrefChange: function(href) {
 				if (!href) {
+					this._date = false;
```
ryantmer: Might make more sense for this to be `null/undefined`? Doesn't make any real difference though, so mostly 🤷‍♂️

m6jones: You are right. Needs to be null can't be undefined or it would run the compute I believe.

PR #38, opened 2018-11-07
Add attest to tests.

m6jones:


package.json
```diff
@@ -10,6 +10,7 @@
     "lint:html": "eslint *.html components/**/*.html test/**/*.{js,html}",
     "lint:wc": "polymer lint -i *.html components/**/*.html",
     "test:polymer:local": "polymer test --skip-plugin sauce",
+    "test:polymer:local:debug": "polymer test --skip-plugin sauce -p",
```
ryantmer: As a heads up, `polymer serve` will give a similar result, the difference being that it just starts a server that can be hit from any browser.

package.json
```diff
@@ -23,6 +24,7 @@
   "author": "D2L Corporation",
   "license": "Apache-2.0",
   "devDependencies": {
+    "@D2L/attest": "^2.5.0",
```
ryantmer: This might be case-sensitive? `@d2l/attest`

m6jones: I think I need to setup authorization to our artifactory?

package.json
m6jones: So that is what I have been doing. Adding this allows you to run a headless test that opens your browser for you. Faster for quick changes.

PR #45, opened 2018-11-14
DE31908: Continue to show the menu drop-down for enrollments in disab…

njostonehouse:
…led courses.

### DIFF
--- a/components/d2l-enrollment-card/d2l-enrollment-card.html
+++ b/components/d2l-enrollment-card/d2l-enrollment-card.html
@@ -67,8 +67,7 @@ Polymer-based web component for a course/enrollment card.
 				cursor: not-allowed;
 			}
 
-			d2l-card[disabled] d2l-enrollment-updates,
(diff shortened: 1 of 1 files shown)


components/d2l-enrollment-card/d2l-enrollment-card.html
```diff
@@ -67,8 +67,7 @@
 				cursor: not-allowed;
 			}
 
-			d2l-card[disabled] d2l-enrollment-updates,
-			d2l-card[disabled] [slot=actions] {
```
njostonehouse: This line prevents actions from appearing on a disabled course. We still want users to be able Pin or Unpin the enrollment (which is the only item available on a disabled course).

m6jones: We may still want to make sure these don't show `d2l-card[disabled] d2l-enrollment-updates` Granted there is really no case where this should happen.

PR #47, opened 2018-11-15
Add tooltip to the enrollments.

m6jones:


.travis.yml
```diff
@@ -4,7 +4,6 @@ addons:
   chrome: stable
 before_script:
 - npm run lint
-- npm run gulp
```
m6jones: running this failed the tests. Don't need to run it in travis. More manual updates to language files.

bower.json
```diff
@@ -32,6 +32,7 @@
     "d2l-organization-hm-behavior": "Brightspace/organization-hm-behavior#^2.0.1",
     "d2l-organizations": "BrightspaceHypermediaComponents/organizations#^0.1.8-hybrid",
     "d2l-status-indicator": "^2.0.0",
+    "d2l-tooltip": "^2.0.4",
```
dbatiste: Note: the version that's needed is `2.0.5` (which should be the version that's picked up as per semver).  I noticed @jstefaniuk-d2l updated tooltips about an hour after this as well, I _think_ for something unrelated.

dbatiste:
Looks like we're still using `fit-content` for the `width`, which I think won't work in Edge and IE11.  I think we should look into/address that if possible.

dbatiste:
Nit: ideally we should update the version of card to the version that's needed to properly support tooltips in the footer.  (`4.1.17`)

PR #49, opened 2018-11-21
DE31983: Bust the cache for the Enrollment so that next time it is re…

njostonehouse:
…trieved the pin state is correct.


njostonehouse:
This seems correct but insufficient to actually fix the cached pin issue, since my courses will still have Enrollments in the wrong spots (and they don't seem to be reloaded to get the new pin states).

njostonehouse:
I don't think so. The tab contents don't even seem to make a new request.

m6jones:
The solution to that would be if `d2l-my-courses` kept each tab separate but in such a way each tab could communicate with other loaded tabs.

We could also reload the tab when selected.

PR #55, opened 2018-12-12

njostonehouse:


njostonehouse:
Seems to have some linting and other path issues. Closing to correct.

PR #57, opened 2018-12-12

njostonehouse:


components/d2l-enrollment-card/d2l-enrollment-card.js
```diff
@@ -0,0 +1,869 @@
...
+				border-radius: 50%;
+				border: 1px solid #f6f7f8;
+				background-color: var(--d2l-color-celestine);
+				position: absolute;
+				top: -0.375rem;
+				right: -0.375rem;
+			}
+
+			/* P2-shadow */
+			:host-context([dir="rtl"]) .d2l-enrollment-card-alert-colour-circle {
```
dbatiste: This style will bleed to LTR in Edge with P3.  This should be removed in a patch.

njostonehouse: Right! Thanks.

PR #60, opened 2018-12-12

njostonehouse:


njostonehouse:
This PR includes merging `hybrid` into `master` as well as the upgrade to Polymer three and the patches.

PR #61, opened 2019-01-10
US102683: Add a d2l- prefix to the npm name to help with future depen…

njostonehouse:
…dency installation.


njostonehouse:
The Polymer 3 upgrade worked fine without the `d2l-` prefix, but the upgrade for `d2l-my-courses` didn't work as well. Changing the package name here to be more consistent with other dependencies, and to hopefully clean up My Courses' upgrade.

njostonehouse:
To avoid causing a mess with things that depend on this hybrid version's name, I think I should just update it in the patch, instead of on the hybrid branch.

PR #62, opened 2019-01-11
US102683: Add a d2l- prefix to the npm name to help with future depen…

njostonehouse:
…dency installation.


njostonehouse:
Patch update: https://github.com/Brightspace/polymer-3-converter/pull/110

PR #65, opened 2019-01-14
Updated the terms.

m6jones:
Some languages completed == complete

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "gulp-rename": "^1.2.3",
     "merge-stream": "^1.0.1",
     "polymer-cli": "^1.7.7",
-    "require-dir": "^1.0.0"
+    "require-dir": "^1.0.0",
+    "del": "^3.0.0"
   }
 }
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -32,6 +32,7 @@
     "gulp-rename": "^1.2.3",
     "merge-stream": "^1.0.1",
     "polymer-cli": "^1.7.7",
-    "require-dir": "^1.0.0"
+    "require-dir": "^1.0.0",
+    "del": "^3.0.0"
```
dbatiste: What was the reason to add `del`?  Seems like there are some dependencies listed here that are not used.  Usually we try to avoid gulp unless absolutely necessary - in the past when we've relied on it, we quickly saw a lot of divergence in project setup and scripts, making things harder to maintain.

m6jones: Gulp is used to build the language files. Del is also part of that script. 
https://github.com/BrightspaceHypermediaComponents/enrollments/blob/master/gulpfile.js

PR #70, opened 2019-02-05
DE32844: Update eslint dependency so that js linting works again.

njostonehouse:


njostonehouse:
Version 4 of this dependency was failing with 
```TypeError: Cannot read property 'define' of undefined
    at Linter.verifyWithSharedScopes```
Updating to 5 fixes it.

PR #86, opened 2019-04-16
US106116-Update activity-usage to use new sdk entity

edwardg8:


EnrollmentEntity.js
```diff
@@ -19,10 +19,44 @@ export class EnrollmentEntity extends Entity {
...
+
+	dueDate() {
+		if (!this._entity) {
+			return;
+		}
+		return this._sirenClassProperty(this._entity, 'due-date');
+	}
+
+	isAttended() {
+		return this._entity && this._entity.hasClass('attended');
```
m6jones: Should we make `'attended'` a constant in the `EnrollmentEntity` 

Unless you think doing that is not a good idea. The idea was that we could document possible classes in the entity file so the file it self is sort of documentation on what the entity can contain.

m6jones: Didn't notice which file we were in.

m6jones: You can do `this.hasClass('attended')` since the parent has that method.

EnrollmentEntity.js
```diff
@@ -19,10 +19,44 @@ export class EnrollmentEntity extends Entity {
...
+	}
+
 	onOrganizationChange(onChange) {
 		const organizationHref = this.organizationHref();
 		// _subEntity builds new sub entity and allows this object to track it.
 		// So all sub entities are dispose when this object is disposed.
 		organizationHref && this._subEntity(OrganizationEntity, organizationHref, onChange);
 	}
+
+	_sirenClassProperty(entity, sirenClass) {
```
m6jones: This is cool!!!

EnrollmentEntity.js
edwardg8: Oh see it now, that's cool, will change this.

PR #97, opened 2019-04-24
Enrollment collection widget skeleton

m6jones:
Demo for enrollment collection widget
fixed self reference to demo data
fixed enrollment card notifications

image

### DIFF
--- a/EnrollmentCollectionEntity.js
+++ b/EnrollmentCollectionEntity.js
@@ -9,6 +9,6 @@ export class EnrollmentCollectionEntity extends Entity {
 			return;
 		}
 		const enrollmentEntities = this._entity.getSubEntitiesByRel(Rels.userEnrollment);
-		return enrollmentEntities.map(e => e.href || undefined).filter(href => href);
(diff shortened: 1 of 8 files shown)


EnrollmentCollectionEntity.js
```diff
@@ -0,0 +1,14 @@
...
+import { Entity } from 'siren-sdk/es6/Entity.js';
+import { Rels } from 'd2l-hypermedia-constants';
+
+export class EnrollmentCollectionEntity extends Entity {
+	enrollmentsHref() {
+		if (!this._entity || !this._entity.getSubEntitiesByRel) {
+			return;
+		}
+		const enrollmentEntities = this._entity.getSubEntitiesByRel(Rels.userEnrollment);
+		return enrollmentEntities.map(e => e.href || undefined).filter(href => href);
```
AdmiralCoco: Minor: I think you can remove the `|| undefined` - If `e.href` is falsy then it will get removed in the `filter`
```suggestion
		return enrollmentEntities.map(e => e.href).filter(href => href);
```

PR #98, opened 2019-04-30
2 enrollment view.

m6jones:
image
When there is two this is how they should look.

### DIFF
--- a/components/d2l-enrollment-collection-widget/d2l-enrollment-collection-widget.js
+++ b/components/d2l-enrollment-collection-widget/d2l-enrollment-collection-widget.js
@@ -33,26 +33,52 @@ class EnrollmentCollectionWidget extends EntityMixin(PolymerElement) {
 					justify-content: space-between;
 				}
 				.decw-flex d2l-enrollment-card {
+					--course-image-height: 136px;
 					flex-grow: 1;
 					margin-top: 30px;
 				}
 				.decw-flex d2l-enrollment-card ~ d2l-enrollment-card {
 					margin-left: 30px;
 				}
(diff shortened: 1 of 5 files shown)


components/d2l-enrollment-collection-widget/d2l-enrollment-collection-widget.js
```diff
@@ -33,26 +33,52 @@ class EnrollmentCollectionWidget extends EntityMixin(PolymerElement) {
...
+								show-unread-discussion-messages
+								show-unread-dropbox-submissions>
+						</d2l-enrollment-card>
+					</template>
+				</div>
+			</template>
+			<template is="dom-if" if="[[_hasTwoEnrollments]]">
+				<div class="decw-flex-two">
+					<d2l-enrollment-hero-banner href="[[_enrollmentHeroHref]]" token="[[token]]"></d2l-enrollment-hero-banner>
+					<d2l-enrollment-card href="[[_enrollmentsHref.0]]" token="[[token]]"
```
AdmiralCoco: What is `.0`?

qwangd2l: It's an array selector for the first element.
I can't seem to find a direct reference to this notation but I found something about it here: https://polymer-library.polymer-project.org/3.0/api/elements/array-selector
```
Note that when dereferencing array indices, the index may be used as a dotted part directly (e.g. users.12.name or ['users', 12, 'name']).
```

m6jones: Yup that is it.

PR #106, opened 2019-05-24
US107194-update references to use new organizations and sdk

edwardg8:


svanherk:
Nice!  When is the entities from here moving into the SDK?

edwardg8:
in next a few PRs, I want to have some cool down time in SDK to no conflict with Maya's. I'll probably do it when I need something from SDK.

svanherk:
Once this merges, we can't update BSI until everywhere that uses the SDK is using version 1

PR #111, opened 2019-05-29
US107197-delete direct use of HM entity of organizations

edwardg8:


edwardg8:
Cannot get rid of `_enrollment` and `_organization` properly until have the event listener removed in my courses content and all courses page, https://github.com/BrightspaceHypermediaComponents/enrollments/blob/master/components/d2l-enrollment-card/d2l-enrollment-card.js#L769

PR #112, opened 2019-05-30
US107197-fix pinning

edwardg8:


svanherk:
Are the other uses of `this._href` correct?  Or shluld they all be `this.href`?  If not, should we have a better name for `this._href`?

edwardg8:
Discovered 4 more places, creating another PR

PR #113, opened 2019-05-30
US107197-use this.href everywhere

edwardg8:
Thanks Stacey replacing the other places too


edwardg8:
I was too excited earlier when discovered the issue.

PR #115, opened 2019-05-31
US107194-use siren action from SDK

edwardg8:


.gitignore
```diff
@@ -1,4 +1,5 @@
 node_modules/
-package-lock.json
+node_modules*/
+.vscode/
 
-node_modules
+package-lock.json
```
ciss1995: Do we want to add package-lock.json back?

edwardg8: no, i'm just re-ordering it, put directory to the top, files to the bottom, nothing is removed

PR #118, opened 2019-06-04
Stop duplications

m6jones:
Basically this makes it smarter about which courses it adds to the list. Don't re-add course and remove courses that no longer exist.

I am tracking courses by index in a separate object. So I can easily maintain the display array.

Only arrays can be loop through. So that is why we have the second flattened array.

### DIFF
--- a/components/d2l-enrollment-summary-view/d2l-enrollment-summary-view.js
+++ b/components/d2l-enrollment-summary-view/d2l-enrollment-summary-view.js
@@ -161,11 +161,11 @@ class D2lEnrollmentSummaryView extends EntityMixin(PolymerElement) {
 				computed: '_computeTags(_courses)'
 			},
 			_title: String,
-			_courses: {
+			_orgHrefs: {
 				type: Array,
 				value: () => []
 			},
-			_coursesIndex: {
+			_orgHrefsByActivitySequence: {
 				type: Object,
 				value: () => {}
 			},
@@ -186,7 +186,7 @@ class D2lEnrollmentSummaryView extends EntityMixin(PolymerElement) {
 		return tags;
 	}
 	_onEnrollmentChange(enrollment) {
-		this._coursesIndex || (this._coursesIndex = {});
+		this._orgHrefsByActivitySequence = this._orgHrefsByActivitySequence || {};
 		enrollment.onOrganizationChange((org) => {
 			this._title = org.name();
 			this._description = org.description();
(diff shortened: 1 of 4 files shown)


components/d2l-enrollment-summary-view/d2l-enrollment-summary-view.js
```diff
@@ -182,13 +186,22 @@ class D2lEnrollmentSummaryView extends EntityMixin(PolymerElement) {
 		return tags;
 	}
 	_onEnrollmentChange(enrollment) {
+		this._coursesIndex || (this._coursesIndex = {});
```
LilyAtD2L: nit: maybe rename the variable `_courseIndex`? It is a bit vague for people who are not familiar with this domain. It will help to have a more explicit naming. For example, if it is to save a mapping courses to the organization links, call it `courseMapToOrglinks`.

components/d2l-enrollment-summary-view/d2l-enrollment-summary-view.js
```diff
@@ -182,13 +186,22 @@ class D2lEnrollmentSummaryView extends EntityMixin(PolymerElement) {
...
 		enrollment.onOrganizationChange((org) => {
-			this._title = org.title();
+			this._title = org.name();
```
LilyAtD2L: Just curious, what is the different between `title` and `name`? We have both for an org?

components/d2l-enrollment-summary-view/d2l-enrollment-summary-view.js
```diff
@@ -182,13 +186,22 @@ class D2lEnrollmentSummaryView extends EntityMixin(PolymerElement) {
...
 			this._description = org.description();
 			org.onSequenceChange((rootSequence) => {
 				rootSequence.onSubSequencesChange((subSequence) => {
 					subSequence.onSequencedActivityChange((activity) => {
-						this._courses = this._courses.concat(activity.organizationHrefs());
+						const orgHrefs = activity.organizationHrefs();
+						const showingOrgHrefs = this._coursesIndex[activity.self()] ? this._coursesIndex[activity.self()] : [];
```
LilyAtD2L: nit: probably simpler to be like the following?
```
const showingOrgHrefs = this._coursesIndex[activity.self()] || []; 
```

components/d2l-enrollment-summary-view/d2l-enrollment-summary-view.js
```diff
@@ -182,13 +186,22 @@ class D2lEnrollmentSummaryView extends EntityMixin(PolymerElement) {
...
+						const addThese = orgHrefs.filter(i => showingOrgHrefs.indexOf(i) < 0);
+						const removeThese = showingOrgHrefs.filter(i => orgHrefs.indexOf(i) < 0);
+
+						let newCourseList = this._courses.concat(addThese);
+						newCourseList = newCourseList.filter(i => removeThese.indexOf(i) < 0);
+						this._courses = newCourseList;
+						this._coursesIndex[activity.self()] = orgHrefs;
```
LilyAtD2L: Looking that the operations here for the arrays, it seems to be `set` might be a better fit here to simplify the implementation?

LilyAtD2L:
Can we add unit tests for `d2l-enrollment-summary-view`?

components/d2l-enrollment-summary-view/d2l-enrollment-summary-view.js
m6jones: title doesn't exists this was a mistake I noticed :(

components/d2l-enrollment-summary-view/d2l-enrollment-summary-view.js
m6jones: What makes `set` better than using a plain object? Asking for more in the Javascript world. (This is me being curious and will be implementing the `set` since in other languages this is 100% the way to go.)

m6jones:
I can add unit test

components/d2l-enrollment-summary-view/d2l-enrollment-summary-view.js
m6jones: IE 11 doesn't like set. Not worth the polyfill in my opinion.

components/d2l-enrollment-summary-view/d2l-enrollment-summary-view.js
LilyAtD2L: 😲

m6jones:
I took it out the entire method to make it easier to unit test. :D 
Reorder the code so now it is easier to read.

PR #121, opened 2019-06-10
Fix Lang term

ciss1995:


components/localize-behavior.js
```diff
@@ -45,7 +45,7 @@ D2L.PolymerBehaviors.Enrollment.LocalizeBehaviorImpl = {
 					'sv': this.sv,
 					'tr': this.tr,
 					'zh': this.zh,
-					'zh-tw': this.zhTw
+					'zh-tw': this.zhtw
```
ciss1995: @m6jones is this the correct change?

PR #125, opened 2019-06-12
Seperate DateAndStatus Mixin

ciss1995:


components/date-text-status-mixin.js
```diff
@@ -2,15 +2,16 @@
...
 /**
  * @polymerMixin
  */
-export const interalDateTextAndStatusMixin = function(superClass) {
-	return class extends mixinBehaviors([
-		D2L.PolymerBehaviors.Enrollment.LocalizeBehavior
-	], superClass) {
-		dateTextAndStatus(isCompletionDate, date) {
+
+export const interalStatusMixin = function(PolymerElement) {
```
edwardg8: i'm not sure if we can do this though, the passed in function can be composite class (with PolymentElement in it), and we just ignore it?

ciss1995: It worked locally so I wanted to give it a shot.

edwardg8: yea, this may only work for the cases where it's directly extending PolymerElement

PR #134, opened 2019-07-08
Adding hidePinning functionality and tests to support Learning Paths.

jlai-d2l:


test/d2l-enrollment-card/d2l-enrollment-card.js
```diff
@@ -169,37 +172,46 @@ describe('d2l-enrollment-card', () => {
...
 	describe('Pinning functionality', () => {
-		it('should have a visible "Unpin" menu item when pinned', () => {
+		it('should have a visible "Unpin" menu item when pinned', done => {
 			component._entity = enrollmentEntity;
-
-			var unpinMenuItem = component.$$('d2l-menu-item.d2l-menu-item-last');
-
-			expect(unpinMenuItem).to.not.be.null;
-			expect(unpinMenuItem.text).to.equal('Unpin');
+			setTimeout(() => {
```
m6jones: Did `afterNextRender` not work?

test/d2l-enrollment-card/d2l-enrollment-card.js
```diff
@@ -169,37 +172,46 @@ describe('d2l-enrollment-card', () => {
...
+				var unpinMenuItem = component.$$('d2l-menu-item.d2l-menu-item-last');
```
dbatiste: Note: the `d2l-menu-item-last` class should be regarded as an internal detail of menus.  I would not assume this class will always be rendered in future version of the component.  It would be better to get the last menu item in a different way.

test/d2l-enrollment-card/d2l-enrollment-card.js
jlai-d2l: No. Unfortunately not. I kept getting inconsistencies between the browser test and the command line test result. In addition, there was already a test using [setTimeout()](https://github.com/BrightspaceHypermediaComponents/enrollments/blob/master/test/d2l-enrollment-card/d2l-enrollment-card.js#L209) and so in addition to the test passing, it also maintains consistency within the file.

PR #135, opened 2019-07-10
Add stickiness to the enrollment summary view progress bar

AllanKerr:
# Changes
1. This pull request adds stickiness to the `d2l-enrollment-summary-view`'s progress bar so that the progress bar stays pinned at the top of the page when scrolling past it.
1. Add a dependency on `lie` to resolve the missing import error during CI runs due to `lie` no longer being brought in implicitly as a sub-dependency.

### DIFF
--- a/components/d2l-enrollment-summary-view/d2l-enrollment-summary-view.js
+++ b/components/d2l-enrollment-summary-view/d2l-enrollment-summary-view.js
@@ -85,6 +85,11 @@ class D2lEnrollmentSummaryView extends EnrollmentsLocalize(EntityMixin(PolymerEl
 					overflow: hidden;
 					padding: 2.45rem 6.75rem 1rem 6.75rem;
 				}
+				.desv-sticky-header {
+					position: sticky;
+					top: 0;
+					z-index: 1000;
+				}
 				.desv-title-bar h1 {
 					@apply --d2l-heading-1;
 					color: var(--d2l-color-ferrite);
@@ -196,6 +201,8 @@ class D2lEnrollmentSummaryView extends EnrollmentsLocalize(EntityMixin(PolymerEl
 					<h1> [[_title]] </h1>
 					<d2l-enrollment-summary-view-tag-list list=[[_tags]]></d2l-enrollment-summary-view-tag-list>
 				</div>
+			</div>
+			<div class="desv-header desv-sticky-header">
 				<d2l-enrollment-summary-view-layout>
 					<div class="desv-progress" slot="first-column">
 						<d2l-enrollment-summary-view-meter
--- a/demo/d2l-enrollment-summary-view/d2l-enrollment-summary-view-demo.html
+++ b/demo/d2l-enrollment-summary-view/d2l-enrollment-summary-view-demo.html
@@ -45,6 +45,7 @@
 				</custom-style>
 			`;
 			document.body.appendChild($_documentContainer.content);
+			window.D2L.Siren.WhitelistBehavior._inTestMode = true;
 		</script>
 	</head>
 	<body class="d2l-typography">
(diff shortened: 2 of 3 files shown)


components/d2l-enrollment-summary-view/d2l-enrollment-summary-view.js
```diff
@@ -271,6 +289,16 @@ class D2lEnrollmentSummaryView extends EnrollmentsLocalize(EntityMixin(PolymerEl
 			'_onEnrollmentChange(_entity)'
 		];
 	}
+
+	connectedCallback() {
+		super.connectedCallback();
+		window.addEventListener('scroll', this._updateNavStickiness);
```
m6jones: Shouldn't this require a `.bind(this)`?

components/d2l-enrollment-summary-view/d2l-enrollment-summary-view.js
```diff
@@ -360,6 +388,20 @@ class D2lEnrollmentSummaryView extends EnrollmentsLocalize(EntityMixin(PolymerEl
...
+		if (!isSticky) {
+			return 0;
+		}
+		const stickyHeader = this.$$('.desv-sticky-header');
+		return stickyHeader.offsetHeight;
+	}
+
+	_updateNavStickiness() {
+		const titleBar = this.$$('.desv-title-bar');
+		const titleBarBottom = titleBar.offsetHeight + titleBar.offsetTop;
```
m6jones: You'll want to use `fastdom.measure` when getting values from the `dom` like this.

components/d2l-enrollment-summary-view/d2l-enrollment-summary-view.js
m6jones: Oops, missed the top...

dbatiste:
Any chance `position:sticky` can be used?  It is not supported by IE11, but this sort of thing is generally really not good (measurements and mutations during scrolling).  The `fastdom` helps a bit, but by no means does it make it amazing.  In some cases we are able to compromise and ignore the lack of IE11 support as long as nothing is breaking.

AllanKerr:
> Any chance `position:sticky` can be used? It is not supported by IE11, but this sort of thing is generally really not good (measurements and mutations during scrolling). The `fastdom` helps a bit, but by no means does it make it amazing. In some cases we are able to compromise and ignore the lack of IE11 support as long as nothing is breaking.

Since it doesn't break any functionality that may be an option, I'll check to make sure we're okay with that.

m6jones:
> > Any chance `position:sticky` can be used? It is not supported by IE11, but this sort of thing is generally really not good (measurements and mutations during scrolling). The `fastdom` helps a bit, but by no means does it make it amazing. In some cases we are able to compromise and ignore the lack of IE11 support as long as nothing is breaking.
> 
> Since it doesn't break any functionality that may be an option, I'll check to make sure we're okay with that.

If we can do this and avoid javascript then awesome! I believe the current method doesn't even work with IE11 anyways which doesn't matter. Sorry, didn't know about the `sticky` position.

dbatiste:
I'm not sure what the design looks like.  As more usage of this sort of behavior is introduced, I think we should be careful.  Some design thought around the usage of sticky (native or otherwise), what happens when there's more than one thing that's sticky, standard vs immersive nav, mobile, etc. would be good.

components/d2l-enrollment-summary-view/d2l-enrollment-summary-view.js
```diff
@@ -191,32 +195,30 @@ class D2lEnrollmentSummaryView extends EnrollmentsLocalize(EntityMixin(PolymerEl
 				}
 			</style>
 
-			<div class="desv-header">
```
AllanKerr: This needed to be flattened since the element with `position: sticky` must be a direct child of the element that it moves within.

PR #139, opened 2019-07-16
Add d2l-enrollment-detail-card loading states

AllanKerr:
# Changes
1. Events
    -  A `d2l-enrollment-detail-card-image-loaded` event is fired when the card image loads.
    - A `d2l-enrollment-detail-card-text-loaded` event is fired when the card text loads.
2. A reveal timeout was added to show the image and text after 2 seconds even if the `--d2l-enrollment-detail-card-loading` style has been applied to force the card to show its content even if the entire page doesn't load successfully.


AllanKerr:
Moving this over to the `d2l-organization-detail-card`

PR #143, opened 2019-07-19
fix flaky CI tests due to race condition.

jlai-d2l:


jlai-d2l:
Will be fixed with @AllanKerr's subsequent PR with additional changes.

AllanKerr:
Re-opening because the pull request that the fix was part of ended getting unexpectedly blocked so these should go into `master`.

PR #144, opened 2019-07-23
d2l-enrollment-hero-banner progress state tree prototype

AllanKerr:


components/StateTree.js
```diff
@@ -0,0 +1,47 @@
+
```
m6jones: I am think maybe we can add this to the `siren-sdk` so I can use this is my other PR. Maybe add a folder in the root called `helpers` This will match with the `brightspaceUI/core` repo

This looks good!!

components/d2l-enrollment-hero-banner/d2l-enrollment-hero-banner.js
```diff
@@ -329,7 +337,21 @@ class EnrollmentHeroBanner extends EnrollmentsLocalize(EntityMixin(PolymerElemen
 			_tags: {
 				type: Array,
 				value: () => ['Due April 25, 2018', '1 hour remaining']
-			}
+			},
+			_orgModulesTree: Object,
+			_orgModules: {
```
m6jones: This is no longer required

AllanKerr:
Added the `StateTree` class as a stop-gap solution to the `siren-sdk`.

PR #145, opened 2019-07-25
Add skeleton loader for the d2l-enrolllment-summary-view to hide the header, side-bar, and detail-cards while loading.

AllanKerr:
Capture

### DIFF
--- a/components/d2l-enrollment-summary-view/d2l-enrollment-summary-view.js
+++ b/components/d2l-enrollment-summary-view/d2l-enrollment-summary-view.js
@@ -70,20 +70,23 @@ class D2lEnrollmentSummaryView extends EnrollmentsLocalize(EntityMixin(PolymerEl
 			</style>
 			<style include="d2l-typography-shared-styles">
 				:host {
+					--desv-header-background-color: #FFFFFF;
 					background: linear-gradient(to bottom, #f6f7f8, #f6f7f8), linear-gradient(to top, #ffffff, #f9fafb);
 					display: block;
 					margin: auto;
 				}
 				.desv-header {
-					background-color: #FFFFFF;
+					background-color: var(--desv-header-background-color);
 					margin: auto;
 				}
 				.desv-title-bar {
+					background-color: inherit;
 					box-sizing: border-box;
 					margin: auto;
 					max-width: 1230px;
 					overflow: hidden;
 					padding: 2.45rem 6.75rem 1rem 6.75rem;
+					position: relative;
 				}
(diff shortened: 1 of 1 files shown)


components/d2l-enrollment-summary-view/d2l-enrollment-summary-view.js
```diff
@@ -195,23 +202,132 @@ class D2lEnrollmentSummaryView extends EnrollmentsLocalize(EntityMixin(PolymerEl
 					}
 				}
 			</style>
+			<!-- Loading Skeleton Styles -->
+			<style>
```
m6jones: So it is standard at d2l to have styles for each `css selector` be in alphabetical order. I suggestion getting an extension to help you do this: such as Alphabetical Sorter.

Not usually how my brain works though since I tend to like to do `width: ...; height: ...;` but it can be useful.

components/d2l-enrollment-summary-view/d2l-enrollment-summary-view.js
```diff
@@ -195,23 +202,132 @@ class D2lEnrollmentSummaryView extends EnrollmentsLocalize(EntityMixin(PolymerEl
...
+					width: 100%;
+					height: 100%;
+
+					padding: inherit;
+					background-color: inherit;
+
+					top: 0;
+					left: 0;
+					bottom: 0;
+					right: 0;
```
m6jones: My above comment would split these up which kinda sucks.

m6jones: @dbatiste Is the reason we don't organize styles like this because it is too hard to keep consistent?

components/d2l-enrollment-summary-view/d2l-enrollment-summary-view.js
AllanKerr: Interesting, that's good to know. I have always grouped and ordered based on semantic relationship to each other but I can certainly switch away from that.

components/d2l-enrollment-summary-view/d2l-enrollment-summary-view.js
AllanKerr: Pasting the previous style @m6jones was referencing before I alphabetized it.
```
position: absolute;
box-sizing: border-box;

width: 100%;
height: 100%;
 
padding: inherit;
background-color: inherit;

top: 0;	
left: 0;
bottom: 0;
right: 0;
```

PR #149, opened 2019-07-30
Add loading states to load the body texts, body images, and header chunks

AllanKerr:
# Changes
1. This pull request adds loading states so that the header text, body text, and body images are loaded as a chunk when all of their sub-pieces have loaded.
1. Force loading has been adding so that sub-pieces will be revealed after 2 seconds even if all sub-pieces in the chunk they belong to haven't been loaded.

### DIFF
--- a/components/d2l-enrollment-summary-view/d2l-enrollment-summary-view.js
+++ b/components/d2l-enrollment-summary-view/d2l-enrollment-summary-view.js
@@ -9,6 +9,7 @@ import './d2l-enrollment-summary-view-layout.js';
 import './d2l-enrollment-summary-view-tag-list.js';
 import './d2l-enrollment-summary-view-meter.js';
 import { EnrollmentsLocalize } from '../EnrollmentsLocalize.js';
+import { afterNextRender } from '@polymer/polymer/lib/utils/render-status';
 
 /**
  * @customElement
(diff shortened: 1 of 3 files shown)


components/d2l-enrollment-summary-view/d2l-enrollment-summary-view.js
```diff
@@ -18,6 +19,8 @@ class D2lEnrollmentSummaryView extends EnrollmentsLocalize(EntityMixin(PolymerEl
 	constructor() {
 		super();
 		this._setEntityType(EnrollmentEntity);
+		this._onOnCourseTextLoaded = this._onOnCourseTextLoaded.bind(this);
```
m6jones: I am assuming on**On**CourseTextLoaded is intentional

test/d2l-enrollment-summary-view/d2l-enrollment-summary-view.js
```diff
@@ -1,9 +1,94 @@
...
+		onCourseOfferingSequenceChangeStub = sandbox.stub();
+		onCourseOfferingSubSequencesChangeStub = sandbox.stub();
+
+		enrollmentEntity = {
+			organizationHref: () => '/organization',
+			onOrganizationChange: onOrganizationChangeStub
+		};
+
+		organizationEntity = {
+			_entity: {
```
m6jones: Why do you need to mock this?

PR #152, opened 2019-08-07
Use grids instead of flex for the my learning

m6jones:
Refactor to use grids instead of a flex. This will make each card have equal space. Still looks and acts the same.

image

### DIFF
--- a/components/d2l-enrollment-collection-widget/d2l-enrollment-collection-widget.js
+++ b/components/d2l-enrollment-collection-widget/d2l-enrollment-collection-widget.js
@@ -28,50 +28,35 @@ class EnrollmentCollectionWidget extends EntityMixin(PolymerElement) {
 				:host {
 					display: block;
 				}
-				.decw-flex {
-					display: flex;
-					justify-content: space-between;
+				.decw-grid {
+					display: grid;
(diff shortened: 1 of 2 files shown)


components/d2l-enrollment-collection-widget/d2l-enrollment-collection-widget.js
```diff
@@ -28,59 +28,44 @@ class EnrollmentCollectionWidget extends EntityMixin(PolymerElement) {
...
+					display: grid;
+					grid-template-columns: [hero-start] 1fr [hero-end];
+					grid-auto-columns: 1fr;
+					grid-gap: 30px 30px;
 				}
-				.decw-flex d2l-enrollment-card {
-					--course-image-height: 136px;
-					flex-grow: 1;
-					margin-top: 30px;
+				.decw-grid-4 {
```
AllanKerr: Are we concerned at all about responsiveness? I could be mistaken but I'm thinking the 3 cards may end up very squished on narrow screens.

m6jones: Yeah, we should be but we don't have a design for it yet. Granted it will probably be pretty obvious on what it should do. I am going to pass it by Colin first though.

PR #159, opened 2019-10-07
DE35575 - Hero Banner title will not overflow white box.

BhamelinD2L:
### Changes
- The css for the hero banner title will now prevent the title from overflowing its container. The container will be the normal height for one or two lines. If the title is greater than two lines, the banner will now wrap the text and expand vertically to fit the title.
- The banner title and content elements are now children of the white box they display in.

Before:
image

After:
image

Excessively long words:
image

### DIFF
--- a/components/d2l-enrollment-hero-banner/d2l-enrollment-hero-banner.js
+++ b/components/d2l-enrollment-hero-banner/d2l-enrollment-hero-banner.js
@@ -88,17 +88,17 @@ class EnrollmentHeroBanner extends EnrollmentsLocalize(EntityMixin(PolymerElemen
 					z-index: -2;
 				}
 				.dehb-info-container {
-					height: 252px;
+					min-height: 252px;
 					position: relative;
 					width: 474px;
 				}
 				.dehb-info-transparent {
 					background: white;
 					border-radius: 8px;
-					height: 100%;
+					height: auto;
 					left: 0;
 					opacity: 0.98;
-					position: absolute;
+					position: relative;
 					top: 0;
 					width: 100%;
 					z-index: -1;
@@ -130,12 +130,12 @@ class EnrollmentHeroBanner extends EnrollmentsLocalize(EntityMixin(PolymerElemen
 					letter-spacing: 0.8px;
 					line-height: 1.4;
 					margin: 0;
+					width: 100%;
 				}
 				.dehb-title {
 					align-items: flex-end;
-					display: inline-block;
-					overflow: hidden; 
-					max-height: 2.8em; /* not a typo meant em */
+					display: flex;
+					overflow: hidden;
 					min-height: 2.26em; /* not a typo meant em */
 				}
 				.dehb-context-menu {
@@ -228,7 +228,8 @@ class EnrollmentHeroBanner extends EnrollmentsLocalize(EntityMixin(PolymerElemen
 						<d2l-organization-image type="wide" href="[[_organizationUrl]]" token=[[token]]></d2l-organization-image>
 					</div>
 					<div class="dehb-info-container">
-						<div class="dehb-info-transparent"><!-- Skeleton for text -->
+						<div class="dehb-info-transparent">
+						<!-- Skeleton for text -->
 						<div class="dehb-base-info-placeholder">
 							<div class="dehb-base-info">
 								<div class="dehb-title">
@@ -247,7 +248,7 @@ class EnrollmentHeroBanner extends EnrollmentsLocalize(EntityMixin(PolymerElemen
 									<div class="dehb-update dehb-update-placeholder"></div>
 								</div>
 							</div>
-						</div>	
+						</div>
 						<div class="dehb-base-info">
 								<div class="dehb-title"><h2>[[_organizationName]]</h2></div>
 								<div class="dehb-context-menu">
--- a/components/d2l-enrollment-hero-banner/d2l-enrollment-hero-banner.js
+++ b/components/d2l-enrollment-hero-banner/d2l-enrollment-hero-banner.js
@@ -2,7 +2,7 @@
 `d2l-enrollment-hero-banner`
 
 Polymer-based web component for a organization name.
- 
+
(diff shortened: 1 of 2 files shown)


components/d2l-enrollment-hero-banner/d2l-enrollment-hero-banner.js
```diff
@@ -247,46 +247,47 @@ class EnrollmentHeroBanner extends EnrollmentsLocalize(EntityMixin(PolymerElemen
 									<div class="dehb-update dehb-update-placeholder"></div>
 								</div>
 							</div>
-						</div>
+						</div>
```
AKobets: There are many lines that have only whitespace  differences. Have you made them on purpose?

components/d2l-enrollment-hero-banner/d2l-enrollment-hero-banner.js
```diff
@@ -227,8 +228,7 @@ class EnrollmentHeroBanner extends EnrollmentsLocalize(EntityMixin(PolymerElemen
 						<d2l-organization-image type="wide" href="[[_organizationUrl]]" token=[[token]]></d2l-organization-image>
 					</div>
 					<div class="dehb-info-container">
-						<div class="dehb-info-transparent"></div>
-						<!-- Skeleton for text -->
+						<div class="dehb-info-transparent"><!-- Skeleton for text -->
```
AKobets: Is it important to change location of this comment?

AKobets:
In the defect https://rally1.rallydev.com/#/39602890252ud/detail/defect/327559170484, screenshot displays  the very long string with no whitespaces. How it looks after the change?

components/d2l-enrollment-hero-banner/d2l-enrollment-hero-banner.js
```diff
@@ -133,7 +133,8 @@ class EnrollmentHeroBanner extends EnrollmentsLocalize(EntityMixin(PolymerElemen
 				}
 				.dehb-title {
 					align-items: flex-end;
-					display: flex;
```
AllanKerr: I don't think we want to be removing `flex` (although I do realize its required for the current solution to function properly). It makes it so the `flex-end` property no longer applies and changes the position of single-line titles so that it no longer matches the design.

m6jones: This is important.

components/d2l-enrollment-hero-banner/d2l-enrollment-hero-banner.js
m6jones: nit: The comment should be on the next line.

components/d2l-enrollment-hero-banner/d2l-enrollment-hero-banner.js
m6jones: Please add an extension that removes trailing whitespace.

components/d2l-enrollment-hero-banner/d2l-enrollment-hero-banner.js
m6jones: Maybe look into what default styles come with `inline-block` and then see which one of those is the one fixing the issue. I am assuming it could be the `width: 100%` but it could be another.

components/d2l-enrollment-hero-banner/d2l-enrollment-hero-banner.js
BhamelinD2L: I've made changes to my environment to take care of this going forward.

components/d2l-enrollment-hero-banner/d2l-enrollment-hero-banner.js
```diff
@@ -247,46 +247,47 @@ class EnrollmentHeroBanner extends EnrollmentsLocalize(EntityMixin(PolymerElemen
...
+									<d2l-enrollment-updates
+										href="[[_organizationUrl]]"
+										token=[[token]]
+										show-unattempted-quizzes
+										show-dropbox-unread-feedback
+										show-ungraded-quiz-attempts
+										show-unread-discussion-messages
+										show-unread-dropbox-submissions>
+									</d2l-enrollment-updates>
+								</div>
```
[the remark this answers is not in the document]
BhamelinD2L: This is the closing div for line 231, and was originally on the same line. None of the text or content was actually contained within the dehb-info-transparent div, which let it overflow vertically and horizontally.

components/d2l-enrollment-hero-banner/d2l-enrollment-hero-banner.js
```diff
@@ -88,17 +88,17 @@ class EnrollmentHeroBanner extends EnrollmentsLocalize(EntityMixin(PolymerElemen
...
 				}
 				.dehb-info-transparent {
 					background: white;
 					border-radius: 8px;
-					height: 100%;
+					height: auto;
 					left: 0;
 					opacity: 0.98;
-					position: absolute;
+					position: relative;
```
AllanKerr: This change seems to cause some issues with the loading skeletons: 
Capture

BhamelinD2L: This is now fixed. 
I have also found that relating to this, the boxes were not quite the original height. They will now match their original height before expanding to fit extra text.

components/d2l-enrollment-hero-banner/d2l-enrollment-hero-banner.js
```diff
@@ -130,11 +131,13 @@ class EnrollmentHeroBanner extends EnrollmentsLocalize(EntityMixin(PolymerElemen
...
 					line-height: 1.4;
 					margin: 0;
+					word-wrap: break-word;
+					width: 100%;
 				}
 				.dehb-title {
 					align-items: flex-end;
 					display: flex;
-					max-height: 2.8em; /* not a typo meant em */
+					overflow: hidden;
```
AllanKerr: nit: I don't this property is used anymore now that we wrap / expand the height.

PR #161, opened 2019-10-24
Enrollment Summary's loaded event will now fire even if no modules exist.

BhamelinD2L:
### Changes
Currently, there is no method to get the total module count from the siren-sdk. Consequently, there's no known time at which we can say all modules have loaded.

The existing solution is to wait 200ms after the first module loads to set the loaded flag to true, and fire the event to show the completion button and text. However, if there are no modules to load, this will never fire and the system will remain in a perpetual loading state.

This has been worked around by introducing a backup delay of 800ms that ensures the show completion continue event will still fire in a reasonable amount of time if no modules are loaded,
with 800ms being enough time for an appropriate module to likely have been found and loaded already, while still being fast enough to meet user expectations.

Before:
image

After:
image

### DIFF
--- a/components/d2l-enrollment-summary-view/d2l-enrollment-summary-view.js
+++ b/components/d2l-enrollment-summary-view/d2l-enrollment-summary-view.js
@@ -628,6 +628,11 @@ class D2lEnrollmentSummaryView extends EnrollmentsLocalize(EntityMixin(PolymerEl
 				}, 200);
 			});
 		});
+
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -42,7 +42,7 @@
     "siren-parser": "^8.0.0",
     "whatwg-fetch": "^3.0.0"
   },
-  "version": "4.0.18",
```
AllanKerr: The `npm` version shouldn't be updated manually. It'll be updated automatically during the automated release stage.

PR #163, opened 2019-12-03
My Learning widget shows an empty state if no enrollments exist

BhamelinD2L:
### Changes
When the My Learning widget has no content to display, it will now display an alert informing the user that there is no content to display, similarly to My Courses.

### Before and after:
image

### DIFF
--- a/components/build/lang/en.js
+++ b/components/build/lang/en.js
@@ -34,6 +34,7 @@ const LangEnImpl = (superClass) => class extends superClass {
 			'inactive': 'Inactive',
 			'moduleProgress': 'Modules',
 			'new': 'New',
+			'noCoursesMessage': 'You don\'t have any courses or learning paths to display.',
 			'overdue': 'Overdue',
 			'pin': 'Pin',
 			'pinActionResult': '{course} has been pinned',
--- a/components/d2l-enrollment-collection-widget/d2l-enrollment-collection-widget.js
+++ b/components/d2l-enrollment-collection-widget/d2l-enrollment-collection-widget.js
@@ -11,12 +11,14 @@ import 'd2l-typography/d2l-typography-shared-styles.js';
 import { EnrollmentCollectionEntity } from 'siren-sdk/src/enrollments/EnrollmentCollectionEntity.js';
 import '../d2l-enrollment-card/d2l-enrollment-card.js';
 import '../d2l-enrollment-hero-banner/d2l-enrollment-hero-banner.js';
+import 'd2l-alert/d2l-alert.js';
+import { EnrollmentsLocalize } from '../EnrollmentsLocalize.js';
 
 /**
  * @customElement
  * @polymer
  */
-class EnrollmentCollectionWidget extends EntityMixin(PolymerElement) {
+class EnrollmentCollectionWidget extends EnrollmentsLocalize(EntityMixin(PolymerElement)) {
 	constructor() {
 		super();
 		this._setEntityType(EnrollmentCollectionEntity);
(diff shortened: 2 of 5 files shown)


components/build/lang/en.js
```diff
@@ -34,6 +34,7 @@ const LangEnImpl = (superClass) => class extends superClass {
 			'inactive': 'Inactive',
 			'moduleProgress': 'Modules',
 			'new': 'New',
+			'noCoursesMessage': 'You don\'t have any courses or learning paths to display.',
```
AKobets: May be message should be more generic - no learning activities to display. Please consult with Maya or Abbie

demo/d2l-enrollment-collection-widget/d2l-enrollment-collection-widget-demo.html
```diff
@@ -64,6 +64,9 @@ <h3>Basic d2l-enrollment-collection-widget demo</h3>
 			<div class="demo">
 				<d2l-enrollment-collection-widget  href="../data/base/enrollmentCollection1.json" token="whatever"></d2l-enrollment-collection-widget>
 			</div>
+			<div class="demo">
+				<d2l-enrollment-collection-widget  href="../data/base/enrollmentCollection4.json" token="whatever"></d2l-enrollment-collection-widget>
```
AKobets: General question: what is `token` attribute for ?

package.json
```diff
@@ -53,6 +53,7 @@
     "@brightspace-ui/core": "^1",
     "@polymer/iron-a11y-announcer": "^3.0.0",
     "@polymer/polymer": "^3.0.0",
+    "d2l-alert": "github:BrightspaceUI/alert",
```
AKobets: should it follow  same convention for version as d2l-button ?

components/build/lang/en.js
AKobets: what is translation process for the langterm. Would it be submitted automatically or you need to follow up with extra work?

BhamelinD2L: I consulted Justin E, Maya, and Abbie to arrive at this message. 
I also confirmed with Maya that the remaining translations will be taken care of automatically as long as the required framework and English translation is in place.

package.json
BhamelinD2L: You're right, I just cleaned this up to conform.

demo/d2l-enrollment-collection-widget/d2l-enrollment-collection-widget-demo.html
BhamelinD2L: I believe it's to authenticate with the back end to retrieve data, but isn't necessary for demos because the back end for the is a collection of .json files.

components/d2l-enrollment-collection-widget/d2l-enrollment-collection-widget.js
```diff
@@ -53,19 +55,26 @@ class EnrollmentCollectionWidget extends EntityMixin(PolymerElement) {
...
+								show-ungraded-quiz-attempts
+								show-unread-discussion-messages
+								show-unread-dropbox-submissions
+								hide-pinning>
+						</d2l-enrollment-card>
+					</template>
+				</div>
+			</template>
+			<template is="dom-if" if="[[!_HasEnrollments]]">
+				<d2l-alert type="call-to-action">
```
AllanKerr: [It looks like `call-to-action` is deprecated.](https://github.com/BrightspaceUI/alert/blob/f0b1031c14d25f4b678aaab64a2932cf92e01f8b/d2l-alert.js#L157) Is there an alternative we can use here?

BhamelinD2L: I missed that, was trying to copy My Courses. `call-to-action` seems to match default alert, so I've removed the parameter to achieve equivalent behavior.

components/d2l-enrollment-collection-widget/d2l-enrollment-collection-widget.js
```diff
@@ -78,6 +87,10 @@ class EnrollmentCollectionWidget extends EntityMixin(PolymerElement) {
 			_countEnrollments: {
 				type: Number,
 				value: 0
+			},
+			_HasEnrollments: {
```
AllanKerr: nit: `has` should be lower case.

PR #167, opened 2019-12-30
US112305 Added load more button, cleaned slot list whitespace, and added homepage href to enrollments

BhamelinD2L:


components/d2l-enrollment-collection-view/d2l-enrollment-collection-view.js
```diff
@@ -24,24 +26,52 @@ class AdminList extends EntityMixinLit(LitElement) {
...
 			enrollment.onOrganizationChange((organization) => {
-				this._items[index].org = organization;
-				this._items[index].href = enrollment.organizationHref();
-				this._items[index].activityUsageUrl = enrollment.userActivityUsageUrl();
-				this.requestUpdate('_items', []);
+				items[index].org = organization;
+				items[index].href = enrollment.organizationHref();
+				items[index].activityUsageUrl = enrollment.userActivityUsageUrl();
+				if (itemsLoadedOnce) {
+					this._items = this._items.concat(items);
```
m6jones: Here you don't need the `concat` since this should only get called if `organization` is already in the list and is being updated individually.

components/d2l-enrollment-collection-view/d2l-enrollment-collection-view.js
```diff
@@ -24,24 +26,52 @@ class AdminList extends EntityMixinLit(LitElement) {
...
+			itemsLoadedOnce = true;
+		});
+	}
+
+	_handleLoadMore() {
+		const nextEnrollmentHref = this._lastEnrollmentCollection.getNextEnrollmentHref();
+		if (nextEnrollmentHref !== null) {
+			dispose(this._lastEnrollmentCollection);
+			if (typeof this._entityType === 'function') {
+				this._lastEnrollmentCollection = entityFactory(this._entityType, nextEnrollmentHref, this.token, entity => {
```
m6jones: Would using `EnrollmentCollectionEntity` be better? Honestly, I don't know. While typing this I am leaning towards `this._entityType` staying.

components/d2l-enrollment-collection-view/d2l-enrollment-collection-view.js
```diff
@@ -93,19 +129,25 @@ class AdminList extends EntityMixinLit(LitElement) {
 					width: 100%;
 					padding-top: 26px;
 				}
+				.d2l-enrollment-collection-view-LoadMore {
+					margin-top:12px;
```
m6jones: I like `margin: N E S W` but lots of people don't. (North East South West) where the vector point to the top of the screen is north.

demo/d2l-enrollment-collection-view/d2l-enrollment-collection-view.html
```diff
@@ -22,7 +22,7 @@
 			<h2>d2l-enrollment-collection-view</h2>
 			<d2l-demo-snippet>
 				<d2l-enrollment-collection-view title-text="My Learning" href="../data/base/enrollmentCollection.json" token="secret"></d2l-enrollment-collection-view>
-			</d2l-demo-snippet>
+			</d2l-demo-snippet>Added
```
m6jones: Why `Added` at the end here?

components/d2l-enrollment-collection-view/d2l-enrollment-collection-view.js
```diff
@@ -82,30 +113,30 @@ class AdminList extends EntityMixinLit(LitElement) {
...
+					margin:12px 12px 12px 0px;
 				}
 				.d2l-enrollment-collection-view-body {
 					box-sizing: border-box;
 					max-width: 1230px;
 					width: 100%;
 					padding-top: 26px;
 				}
+				.d2l-enrollment-collection-view-load-more {
+					margin:12px 12px 12px 0px;
```
dbatiste: I think we will want to have consistent spacing. We could provide a slot for paging that provides the consistent layout.  Also, is the button meant to be 100% width or no?  Probably another things we should be consistent with.

dbatiste: Nit: a matter of convention/practice - space after the `:`, and also the `0px` value should just be `0`.  The value is a micro-optimization from the past.  So this would be `margin: 12px 12px 12px 0;`

PR #170, opened 2020-01-09
US112310: Make the My Learning widget responsive

AllanKerr:
# Changes

- Define the number of columns based on the container width and a column width of 350 px to match My Courses. This was done with `offsetWidth` instead of media queries because different homepage widget layouts result in My Learning having different widths relative to the page size.
- Conditionally show the hero banner based on the number of columns and the number of enrollments.

# Screenshots

### DIFF
--- a/components/d2l-enrollment-collection-widget/d2l-enrollment-collection-widget.js
+++ b/components/d2l-enrollment-collection-widget/d2l-enrollment-collection-widget.js
@@ -13,6 +13,7 @@ import '../d2l-enrollment-card/d2l-enrollment-card.js';
 import '../d2l-enrollment-hero-banner/d2l-enrollment-hero-banner.js';
 import 'd2l-alert/d2l-alert.js';
 import { EnrollmentsLocalize } from '../EnrollmentsLocalize.js';
+import { ResizeObserver } from 'd2l-resize-aware/resize-observer-module.js';
 
 /**
  * @customElement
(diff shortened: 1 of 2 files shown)


components/d2l-enrollment-collection-widget/d2l-enrollment-collection-widget.js
```diff
@@ -22,6 +22,22 @@ class EnrollmentCollectionWidget extends EnrollmentsLocalize(EntityMixin(Polymer
 	constructor() {
 		super();
 		this._setEntityType(EnrollmentCollectionEntity);
+		this._onResize = this._onResize.bind(this);
+	}
+
+	connectedCallback() {
+		super.connectedCallback();
+		window.addEventListener('resize', this._onResize);
```
dbatiste: I think the resize observer (on the root container) would be a preferred way to handle this.  Any chance of trying that out if not already?  We have an import for it.  

One advantage to it is that will only be triggered when the container resizes, and the dimensions will be provided thereby avoiding an explicit measurement. The polyfill will fallback to resize event for any browser that does not support it (Edge and Safari... but Safari supports in TP, and Edge will get it with CrEdge).

AllanKerr: Switched to use the `ResizeObserver`, thanks for the tip!

PR #171, opened 2020-01-09
US112310: Give the My Learning widget a basic loading spinner

AllanKerr:
# Changes
- Add a basic loading spinner that is reveals the widget when all sub-entities have loaded. This implementation does not wait until all images are loaded since their loading is handled by events rather than loading them as sub-entities. Image support will be added in the future as required.


components/d2l-enrollment-collection-widget/d2l-enrollment-collection-widget.js
```diff
@@ -87,34 +87,51 @@ class EnrollmentCollectionWidget extends EnrollmentsLocalize(EntityMixin(Polymer
...
+									hide-pinning>
+							</d2l-enrollment-card>
+						</template>
+					</div>
+				</template>
+				<template is="dom-if" if="[[!_hasEnrollments]]">
+					<d2l-alert>
+						[[localize('noCoursesMessage')]]
+					</d2l-alert>
+				</template>
```
AllanKerr: This chunk has not changed aside from being indented due to being placed in a `dom-if` block.

PR #173, opened 2020-01-15
My Learning Page

AllanKerr:
# Changes

This pull request finalizes the My Learning page by adding:
- Load more paging
- Searching
- Loading spinners
- Responsiveness
- Localization
- Loading Skeletons

### DIFF
--- a/components/d2l-enrollment-collection-view/d2l-enrollment-collection-view.js
+++ b/components/d2l-enrollment-collection-view/d2l-enrollment-collection-view.js
@@ -24,6 +24,7 @@ class AdminList extends LocalizeMixin(EntityMixinLit(LitElement)) {
 		super();
 		this._items = [];
 		this._loadedImages = [];
+		this._hasFirstLoad = false;
 		this._mainPageLoad = new Promise(() => {});
 		this._setEntityType(EnrollmentCollectionEntity);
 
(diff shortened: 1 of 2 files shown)


components/d2l-enrollment-collection-view/d2l-enrollment-collection-view.js
```diff
@@ -17,14 +18,30 @@ import '@brightspace-ui/core/components/list/list-item-content.js';
 import '@brightspace-ui/core/components/inputs/input-search.js';
 import 'd2l-loading-spinner/d2l-loading-spinner.js';
```
dbatiste: Should be able to pull in loading-spinner from `core`.

AllanKerr: Fixed, thanks!

dbatiste:
I know we've spoken briefly about this previously, but I think we should create a couple work items to for standardizing/integrating search, paging/load-more, and empty messages.  The empty messages is on Jeff's radar and we'll create a shared component for it once design is ready.

m6jones:
That would be amazing. There some very early prototype patterns on how we can code it up here and the other learning path things. But a lot of this is hidden away in the component because we want more thought and effort into making any pattern. 

Both of us are not too proud of how the code turned out but the results on the page are pretty awesome.

AllanKerr:
@dbatiste I completely agree, way too much code is being duplicated for implementing paging. It definitely needs to be standardized. I'd like to reach a point where all of it is standardized so paging is as easy as adding a load-more, paged, search component and passing it a URL.

PR #182, opened 2020-01-22
Fixed floating dot in My Learning metadata

zinabat:
[DE37487](https://rally1.rallydev.com/#/detail/defect/362507044272?fdp=true)

Floating dot was appearing in the metadata for My Learning in cases where a course was partially completed. This was due to siren-sdk returning undefined instead of null.
Screen Shot 2020-01-22 at 1 11 00 PM

The floating dot will no longer render in these cases.

### DIFF
--- a/components/d2l-enrollment-collection-view/d2l-enrollment-collection-view.js
+++ b/components/d2l-enrollment-collection-view/d2l-enrollment-collection-view.js
@@ -535,8 +535,8 @@ ${
 			});
 
 			enrollment.onUserActivityUsageChange((activityUsage) => {
-				// activityUsage.date() is undefined for incomplete courses, so using == here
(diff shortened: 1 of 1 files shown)


components/d2l-enrollment-collection-view/d2l-enrollment-collection-view.js
```diff
@@ -535,7 +535,8 @@ ${
 			});
 
 			enrollment.onUserActivityUsageChange((activityUsage) => {
-				items[index].hasDueDate = items[index].hasDueDate || activityUsage.date() !== null;
+				const date = activityUsage.date();
+				items[index].hasDueDate = items[index].hasDueDate || (date !== null && date !== undefined);
```
AllanKerr: One thing that we could do here is use `!=` then Javascript will coerce the value to `false` whether it is `null` or `undefine`. (assuming the linter doesn't complain)

PR #184, opened 2020-01-22
Match the total number

iFun:
[DE37431](https://rally1.rallydev.com/#/39602890252d/detail/defect/361868532492?fdp=true)
Activity count in Herobanner is inconsistent compared to the activity count in the LP Summary page


components/d2l-enrollment-hero-banner/d2l-enrollment-hero-banner.js
```diff
@@ -516,22 +516,21 @@ class EnrollmentHeroBanner extends DateTextAndStatusMixin(EnrollmentsLocalize(En
 						sequencedActivityNode.removeAllChildren();
 
 						const orgKey = `${subSequence.index()}-${sequencedActivity.index()}`;
-						this._updateCourseOfferingModules(orgKey, organizationEntity, sequencedActivityNode);
+						this._updateCourseOfferingModules(orgKey, organizationEntity, sequencedActivityNode, sequencedActivity.index() + 1);
```
AllanKerr: We shouldn't be using the index like this. The index is only use to keep track of the items position within the tree. You should be either counting the number of sequenced activities here and passing that value down or computing it from the number of `moduleItems` in `_updateOrganizationCompletion`. If the items you need aren't there then you can call `modulesTree.setChild(...)` to pass additional items to the `_updateOrganizationCompletion` call.

zinabat:
I'm still not seeing a matching activity count when I checkout your branch while testing with the LE - the activities count isn't being updated properly for some reason.
Screen Shot 2020-01-22 at 4 07 17 PM
Screen Shot 2020-01-22 at 4 07 59 PM

iFun:
> I'm still not seeing a matching activity count when I checkout your branch while testing with the LE - the activities count isn't being updated properly for some reason.
> Screen Shot 2020-01-22 at 4 07 17 PM
> Screen Shot 2020-01-22 at 4 07 59 PM

thank you working on a fix now

iFun:
@AllanKerr do you have time to take another look?

components/d2l-enrollment-hero-banner/d2l-enrollment-hero-banner.js
```diff
@@ -609,16 +607,19 @@ class EnrollmentHeroBanner extends DateTextAndStatusMixin(EnrollmentsLocalize(En
...
+		let total = 0;
 		moduleItems.forEach(item => {
-			const curCompletion = item.orgModule.completion() || {};
-			const prevCompletion = completions[item.orgKey] || {};
-
-			completions[item.orgKey] = {
-				completed: (prevCompletion.completed || 0) + (curCompletion.completed || 0),
-				total: (prevCompletion.total || 0) + (curCompletion.total || 0)
-			};
+			if (typeof item.hasClass === 'function' && item.hasClass('sequenced-activity')) {
```
AllanKerr: nit: Not a huge fan of having to check for a `hasClass` function, would prefer if we attached something like a type property to both the modules and activity objects that are added.

iFun: sure i can add a type for that

PR #202, opened 2020-02-27
added return url for continue button

iFun:
wait for confirmation this is the correct way to do it


iFun:
closed due to different approach

PR #204, opened 2020-03-09
Disabling link if course is inactive

zinabat:
[DE37652](https://rally1.rallydev.com/#/detail/defect/365729624044?fdp=true)

Ensures an activity title does not appear blue (clickable) on the View Learning Path page if it is inactive.

Screen Shot 2020-03-18 at 11 07 23 AM

**Quality**
Tested with demo pages and local BSI


zinabat:
Yes, that's a good idea.

PR #206, opened 2020-03-26
use hasDate instead of hasDueDate

maboelsoudd2l:
before:
image

after:
image

### DIFF
--- a/components/d2l-enrollment-collection-view/d2l-enrollment-collection-view.js
+++ b/components/d2l-enrollment-collection-view/d2l-enrollment-collection-view.js
@@ -535,14 +535,14 @@ ${
 				items[index].href = organizationHref;
 				items[index].activityUsageUrl = enrollment.userActivityUsageUrl();
 				items[index].hasDueDate = items[index].hasDueDate || organization.endDate() !== null;
-				items[index].startDate = organization.startDate() !== null;
-				items[index].hasDate = items[index].hasDueDate || items[index].startDate;
+				items[index].hasStartDate = organization.startDate() !== null;
+				items[index].hasDate = items[index].hasDueDate || items[index].hasStartDate;
 			});
 
 			enrollment.onUserActivityUsageChange((activityUsage) => {
 				const date = activityUsage.date();
 				items[index].hasDueDate = items[index].hasDueDate || (date !== null && date !== undefined);
-				items[index].hasDate = items[index].hasDueDate || items[index].startDate;
+				items[index].hasDate = items[index].hasDueDate || items[index].hasStartDate;
 			});
 		});
 		await enrollmentCollection.subEntitiesLoaded();
--- a/components/d2l-enrollment-collection-view/d2l-enrollment-collection-view.js
+++ b/components/d2l-enrollment-collection-view/d2l-enrollment-collection-view.js
@@ -363,9 +363,9 @@ class EnrollmentCollectionView extends LocalizeMixin(EntityMixinLit(LitElement))
 						<d2l-enrollment-summary-view-tag-slot-list>
 							<span slot="first">${this._enrollmentType(item)}</span>
 ${
-	item.hasDueDate ? html`
+	item.hasDate ? html`
 		<d2l-user-activity-usage slot="middle" href=${ifDefined(item.activityUsageUrl)} .token=${this.token}>
-			<d2l-organization-date slot="default" href=${item.href} .token=${this.token} hide-course-start-date
+			<d2l-organization-date slot="default" href=${item.href} .token=${this.token}
 			></d2l-organization-date>
 		</d2l-user-activity-usage>
 	` : null
@@ -535,11 +535,14 @@ ${
 				items[index].href = organizationHref;
 				items[index].activityUsageUrl = enrollment.userActivityUsageUrl();
 				items[index].hasDueDate = items[index].hasDueDate || organization.endDate() !== null;
+				items[index].startDate = organization.startDate() !== null;
+				items[index].hasDate = items[index].hasDueDate || items[index].startDate;
 			});
 
 			enrollment.onUserActivityUsageChange((activityUsage) => {
 				const date = activityUsage.date();
 				items[index].hasDueDate = items[index].hasDueDate || (date !== null && date !== undefined);
+				items[index].hasDate = items[index].hasDueDate || items[index].startDate;
 			});
 		});
 		await enrollmentCollection.subEntitiesLoaded();


components/d2l-enrollment-collection-view/d2l-enrollment-collection-view.js
```diff
@@ -535,11 +535,14 @@ ${
 				items[index].href = organizationHref;
 				items[index].activityUsageUrl = enrollment.userActivityUsageUrl();
 				items[index].hasDueDate = items[index].hasDueDate || organization.endDate() !== null;
+				items[index].startDate = organization.startDate() !== null;
```
AKobets: on line 538 should it be `items[index].hasStartDate = ...`

maboelsoudd2l: ooopsies

AKobets: it is easy to make mistake if you have both variable and function with same name

AKobets:
@maboelsoudd2l  could you please provide screenshot with LPath that has courses: 1) has both Start Date and End Date, 2) has both Start Date and Due data, 3) has both Due Date and End Date, 4) has three dates Start Date, Due Date and  End Date

maboelsoudd2l:
@AKobets i don't know how to assign due date on a course to a user (nor have i seen it so far on UAT). If this is a use-case that we need to test, can you give me more info on how to set it up?

AKobets:
> @AKobets i don't know how to assign due date on a course to a user (nor have i seen it so far on UAT). If this is a use-case that we need to test, can you give me more info on how to set it up?

due dates are assigned by als. So it could be done if user is assigned course via learning groups with completion tracking on and due date specified. Lets have screenshots for both start and end date first and I will try to setup course with assigned due date.

maboelsoudd2l:
I already have screenshots for both start and end, at the PR description

AKobets:
> I already have screenshots for both start and end, at the PR description

I do see only one date displayed in the course card

maboelsoudd2l:
That's apparently what design wants, they don't want both start and end to show up at the same time.

AKobets:
activity usages due dates are confusing. They are used in enrollments collection view, however it looks like due date will never be displayed to user because d2l-organization-date is not aware of them, same as Siren SDK organization entity. 
See https://github.com/BrightspaceHypermediaComponents/siren-sdk/blob/58bff8ea80e62bddcbef7de25045679b21c86214/src/organizations/OrganizationEntity.js#L51

Regarding mix of start/due dates - d2l-organization-date will display dates according to logic in OrganizationEntity.processedDate  and it was probably tested well. 

Testing the case with activity usages date not null and Start & End dates are null is out of scope for this PR. I will test it by my own.

components/d2l-enrollment-collection-view/d2l-enrollment-collection-view.js
```diff
@@ -535,11 +535,14 @@ ${
...
 				items[index].activityUsageUrl = enrollment.userActivityUsageUrl();
 				items[index].hasDueDate = items[index].hasDueDate || organization.endDate() !== null;
+				items[index].hasStartDate = organization.startDate() !== null;
+				items[index].hasDate = items[index].hasDueDate || items[index].hasStartDate;
 			});
 
 			enrollment.onUserActivityUsageChange((activityUsage) => {
 				const date = activityUsage.date();
 				items[index].hasDueDate = items[index].hasDueDate || (date !== null && date !== undefined);
+				items[index].hasDate = items[index].hasDueDate || items[index].hasStartDate;
```
AKobets: Why items[index].hasDate needs to be changed here? and why do you need to to OR it with items[index].hasStartDate which is assigned in another handler? If hasDueDate is true, it is probably enough to set items[index].hasDate to true

AKobets: I'm new to this code and may be misunderstand  how callbacks passed to onOrganizationChange and onUserActivityUsageChange will be executed. If they always executed syncroneously and in the same order 1st onOrganizationChange  and then onUserActivityUsageChange , then there is no risk of overriding items[index].hasDate . 
If those 2 callbacks could be executed in any order then probably lines 539 and 545 should start with `items[index].hasDate = items[index].hasDate || ...`

maboelsoudd2l: because redundancy. I just wanna make sure.

maboelsoudd2l: it seemed to me that theres a chance that `onUserActivityUsageChange` might cause the date to not show up. So i would rather check instead of a month later we find out a new bug due to some edge case.

maboelsoudd2l: @AKobets thoughts?

maboelsoudd2l:
> activity usages due dates are confusing

I agree it is confusing, according to what I have been told by @m6jones, d2l-organization-date is supposed to be inside of d2l-user-activity-usage. I think the logic is that they access different hypermedia entities, but I think the end result is that it complicates code and leaves room for bugs.

Personally I think the Ideal should be the following:
- some kind of `course-date-manager` where we feed it the logic of how dates show up.
- we rename `d2l-organization-date` to `d2l-course-date` for clarity (also give it properties of `start-date` and `end-date`)
- rename `d2l-user-activity-usage` to `d2l-course-due-date` and remove the cascading

>  d2l-organization-date will display dates according to logic

My personal opinion it seems like that logic might be a mystery 😄

maboelsoudd2l:
@AKobets what do we need in order to ship this PR?

components/d2l-enrollment-collection-view/d2l-enrollment-collection-view.js
AKobets: > @AKobets thoughts?

I'm OK with this approach.

components/d2l-enrollment-collection-view/d2l-enrollment-collection-view.js
```diff
@@ -535,11 +535,14 @@ ${
 				items[index].href = organizationHref;
 				items[index].activityUsageUrl = enrollment.userActivityUsageUrl();
 				items[index].hasDueDate = items[index].hasDueDate || organization.endDate() !== null;
+				items[index].hasStartDate = organization.startDate() !== null;
+				items[index].hasDate = items[index].hasDueDate || items[index].hasStartDate;
 			});
```
zinabat: I appreciate descriptive boolean properties like this - easy to read

PR #208, opened 2020-03-30
incorporate d2l-button

maboelsoudd2l:
[DE38478](https://rally1.rallydev.com/#/detail/defect/379552725536?fdp=true)

Incorporated d2l-button, because it seems like there's a lot of css work being done to achieve the same result, which is unnecessary.

before:
image

after:
image

### DIFF
--- a/components/d2l-enrollment-summary-view/d2l-enrollment-summary-view.js
+++ b/components/d2l-enrollment-summary-view/d2l-enrollment-summary-view.js
@@ -324,6 +324,8 @@ class D2lEnrollmentSummaryView extends EnrollmentsLocalize(EntityMixin(PolymerEl
 							primary
 							disabled$="[[!_continueModule.href]]"
 							href$="[[_continueModule.href]]"
+							on-click="_navigateContinueModule"
+							role="link"
 							aria-label$="[[localize('continueToModule', 'module', _continueModule.title)]]">
 								[[localize('continue')]]
 						</d2l-button>
@@ -501,6 +503,15 @@ class D2lEnrollmentSummaryView extends EnrollmentsLocalize(EntityMixin(PolymerEl
 			tags.push(courses.length > 1 ? courses.length + ' Activities' : '1 Activity');
 		}
 		return tags;
+	}
+	_navigateContinueModule() {
+		this._setWindowLocationHref(this._continueModule.href);
+	}
+
+	_setWindowLocationHref(href) {
+		if (href !== '') {
+			window.location.href = href;
+		}
 	}
 	_onCoursesChange(courses) {
 		const coursesLoaded = courses.reduce((map, href) => {
--- a/components/d2l-enrollment-summary-view/d2l-enrollment-summary-view.js
+++ b/components/d2l-enrollment-summary-view/d2l-enrollment-summary-view.js
@@ -329,7 +329,7 @@ class D2lEnrollmentSummaryView extends EnrollmentsLocalize(EntityMixin(PolymerEl
 							aria-label$="[[localize('continueToModule', 'module', _continueModule.title)]]">
 								[[localize('continue')]]
 						</d2l-button>
-						<span>[[_continueModule.title]]</span>
+						<span role="presentation" aria-hidden="true">[[_continueModule.title]]</span>
 					</div>
 				</d2l-enrollment-summary-view-layout>
 			</div>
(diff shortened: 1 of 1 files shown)


maboelsoudd2l:
Huge caveat: The progress bar doesn't look great in mobile, but that's not because of me. Looking at the css, there seems to be wonky stuff going on unecessarily (like position: absolute).
image

components/d2l-enrollment-summary-view/d2l-enrollment-summary-view.js
```diff
@@ -359,14 +320,13 @@ class D2lEnrollmentSummaryView extends EnrollmentsLocalize(EntityMixin(PolymerEl
 							<div class="desv-placeholder desv-button-placeholder"></div>
 							<div class="desv-placeholder desv-compact-text-placeholder desv-continue-title-placeholder"></div>
 						</div>
-						<a
-							class="desv-button"
+						<d2l-button
```
m6jones: This isn't a button! that is why it was a link with a button style.

m6jones: You can see if role="link" will fix it I rather use the `d2l-button` if we can.

maboelsoudd2l:
I have a feeling that if we wrapped an `<a>` tag around a button will make us fail WCAG 2.1 rule: https://www.w3.org/WAI/WCAG21/quickref/?versions=2.1#qr-navigation-mechanisms-refs

also design guidelines tells us against doing that here http://design.d2l/components/buttons/ (scroll to "don't". its against using buttons for navigation)

m6jones:
> I have a feeling that if we wrapped an `<a>` tag around a button will make us fail WCAG 2.1 rule: https://www.w3.org/WAI/WCAG21/quickref/?versions=2.1#qr-navigation-mechanisms-refs
> 
> also design guidelines tells us against doing that here http://design.d2l/components/buttons/ (scroll to "don't". its against using buttons for navigation)

Design wants it to look like a button. But for accessibility. It should read and act as a link. I am happy if we can use the `d2l-button` component or the styles defined in core.

maboelsoudd2l:
FYI @dbatiste

zinabat:
I also don't know that wrapping with an anchor tag is advisable, but it is something that currently exists in the widget to be aware of. Ideally the button would see the href and know to treat it as a link.

maboelsoudd2l:
@zinabat I was talking to @dbatiste about this, wrapping anchor tags around buttons is invalid HTML. According to official spec (https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element) the `<a>` cannot wrap any interactive content. A button falls under that description. 

Even though it is invalid, it seems like browsers render it the way want it to. I tested it on screen readers and it seems like it works for NVDA and VoiceOver. Even tabbing kinda works. So since i don't see functionality broken, I'm not in a hurry to make changes. What I'm leaning towards is waiting until design gets back and lets us know if they recommend button or link-button or link.

maboelsoudd2l:
stole `_setWindowLocationHref` from Quick-Eval.

added a `role="link"` that makes

m6jones:
Going to test it

maboelsoudd2l:
Thanks @m6jones , based on my testing, I got: "link, continue by Continue by navigating to module, Accessible Word Document"

components/d2l-enrollment-summary-view/d2l-enrollment-summary-view.js
```diff
@@ -359,14 +320,15 @@ class D2lEnrollmentSummaryView extends EnrollmentsLocalize(EntityMixin(PolymerEl
...
 							primary
 							disabled$="[[!_continueModule.href]]"
 							href$="[[_continueModule.href]]"
+							on-click="_navigateContinueModule"
+							role="link"
 							aria-label$="[[localize('continueToModule', 'module', _continueModule.title)]]">
 								[[localize('continue')]]
-						</a>
+						</d2l-button>
 						<span>[[_continueModule.title]]</span>
```
m6jones, line 332: Hide this from the screen reader too! no reason to duplicate it.

m6jones:
Also, for the actual meat of the test. VoiceOver works! @maboelsoudd2l killed it! Nice!

components/d2l-enrollment-summary-view/d2l-enrollment-summary-view.js
maboelsoudd2l: hide what? the `_continueModule.title` ?

m6jones: The span, yeah from screen readers.

m6jones: Right now they get to the button and the name is told to them and they can act on it.
The next thing they hear is the name again but cannot act on it.

m6jones: `role="presentation"` `aria-hidden="true"` attributes on the span should work.

maboelsoudd2l: awesome stuff! 👍

PR #209, opened 2020-03-30
added a color override for small-text to ferrite

maboelsoudd2l:
i highly recommend we don't do this. I recommend we update d2l-typography. That way we don't hard code this text to ferrite. And also if d2l-typography or d2l-colors change, we can inherit these changes.

fixes: https://rally1.rallydev.com/#/357252966636d/detail/defect/366100715660

The logic for changing the color is that tungsten is no longer a sufficient contrast level. I definitely think that this applies to all small text and thats why i think a change should be made across the board.

Quality
Tested with demo pages and local BSI


maboelsoudd2l:
before: 
image

after:
image

dbatiste:
It looks like the current color passes (WCAG 2.1 AA) -> [contrast for tungsten](https://webaim.org/resources/contrastchecker/?fcolor=6E7376&bcolor=FFFFFF).

The design for small body text still indicates tungsten for the color.  I mentioned this to Jeff G. and he's added it to his list of core things to look at further since he agrees it should be looked at further.  Any fix we make to the color should be made in core as you suggested.

m6jones:
Would @maboelsoudd2l be able to make this change in core?

dbatiste:
We definitely welcome such changes but they should be done in conjunction with a Designer who updates design.d2l and any other design assets they have.

PR #215, opened 2020-05-01
DE38903 - Fix closed badge

svanherk:
Related to BrightspaceHypermediaComponents/siren-sdk#184, I'm migrating usages of the `processedDate` function from `OrganizationEntity` to no longer access `beforeStartDate` and `afterEndDate` directly, and to instead use the new functions.

In this case, the `d2l-enrollment-card` and `d2l-enrollment-hero-banner` were not getting the closed badge applied if `hideCourseEndDate` was true, even though that setting should only affect the date text displayed.  Using the new functions fixes that.  The logic between the two is exactly the same, as are the tests.

### DIFF
--- a/test/d2l-enrollment-hero-banner/d2l-enrollment-hero-banner.js
+++ b/test/d2l-enrollment-hero-banner/d2l-enrollment-hero-banner.js
@@ -23,6 +23,8 @@ describe('d2l-enrollment-hero-banner', () => {
 		dateStub,
 		isActiveStub,
 		processedDateStub,
+		isBeforeStartDateStub,
+		isAfterEndDateStub,
 		date;
 
 	beforeEach(() => {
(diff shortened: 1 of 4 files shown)


test/d2l-enrollment-hero-banner/d2l-enrollment-hero-banner.js
```diff
@@ -394,26 +402,26 @@ describe('d2l-enrollment-hero-banner', () => {
 					component._setEnrollmentStatus('completed');
 				};
 				var setBeforeStart = function() {
-					component._setOrganizationDate({
```
svanherk, line 397: This feels like it was just wrong and not properly setting the `isBeforeStartDate`, so following what `d2l-enrollment-card` tests were doing

PR #219, opened 2020-07-13
Sequences 3

ChrisKraljevicD2L:
There's a new major version of sequences available. Nothing about the old code broke but there is a substantial amount of new code brought over from a different repository. There are a few new dependencies for the new code.

I'm bumping this here so BSI's `package-lock.json` doesn't get messy.


AllanKerr:
Is there a reason we needed a major version bump if there aren't any breaking changes? Just want to make sure I'm not missing anything before approving.

ChrisKraljevicD2L:
Good point. I think I'll re-release this as a `2.x` release instead. Sorry about that.

PR #220, opened 2020-07-20
fix: Localization is broken for enrollment collection in 20.20.8 for Edge

zinabat:
closes https://rally1.rallydev.com/#/detail/defect/405077696728?fdp=true

Defect caused by a switch to `rollup` in BSI builds

## Changes
- Changed so that localization imports are explicit
- Installed `es-dev-server` because we should start moving away from `polymer-cli`
  - This will start a dev server and auto-watch files on `npm start`, just like core

## Quality
- Tested that all of the demo pages work. Performed some light maintenance on them
- Tested that My Learning page works in Legacy Edge with local LMS

### DIFF
--- a/demo/d2l-user-activity-usage/d2l-user-activity-usage-demo.html
+++ b/demo/d2l-user-activity-usage/d2l-user-activity-usage-demo.html
@@ -4,12 +4,11 @@
 		<meta charset="utf-8">
 		<meta name="viewport" content="width=device-width, minimum-scale=1, initial-scale=1, user-scalable=yes">
 		<title>d2l-user-activity-usage</title>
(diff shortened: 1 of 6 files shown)


demo/d2l-user-activity-usage/d2l-user-activity-usage-demo.html
```diff
@@ -1,126 +1,125 @@
...
-				</template>
-            </demo-snippet>
-
-		</div>`;
-
-document.body.appendChild($_documentContainer.content);
-		</script>
-	</body>
-</html>
+<!doctype html>
```
zinabat, line 1: This is just house keeping, removed non-resolving duplicate imports. Not sure why the diff is the whole file.

PR #223, opened 2020-11-24
switch to GitHub actions

dlockhart:


.travis.yml
```diff
@@ -1,31 +0,0 @@
...
-deploy:
-  provider: releases
-  api_key: "$GITHUB_RELEASE_TOKEN"
-  on:
-    tags: true
-env:
-  global:
-  - OWNER_NAME: Brightspacehypermediacomponents
-  - REPO_NAME: enrollments
-  - DEFAULT_INCREMENT: patch
```
svanherk, line 28: This is one of those repos where people will be use to the automatic releasing each time they merge.  It's not an app repo, but it's probably considered similar to `activities` (which _does_ have apps in it) and `organizations`, which they will also be used to the auto-releasing

svanherk: Not saying we can't change it, but those three (and maybe `sequences`...) will probably cause some confusion and we'd probably want to highlight in #web-platform

dlockhart: Yeah in this case I checked with @m6jones and she was good with the switch. For the others I'm going to need to start some team threads in Slack I think!

svanherk (approved):

PR #233, opened 2021-05-31
add activityCount to fr-on

lloydwaddell:
```Context```
FR-ON was missing the new lang term for activity count. Term was copied from fr.json file.


lloydwaddell:
no longer needed this was added by the telo serge pr

PR #238, opened 2021-08-11
DE444450 - Convert the processedDate to the user's LMS set timezone

PaulKrznaric:
This is an HTML change, it doesn't affect how the data is displayed, but is found in the source HTML as a value.


jordankay13 (approved):

klhuff:
What the, where did all the comments go.

klhuff (approved):

PaulKrznaric:
@klhuff One is for the HTML layer only, one is for the visual display

PR #244, opened 2021-12-16
feat: switch various components to use `@brightspace-ui/core`

devpow112:
All of these were simply drop in replacements as the files from the individual packages simply had a single line that was an import from `@brightspace-ui/core`. As part of the `d2l-icons` removal I also trimmed all the `d2l-` off the icon names as this is actually not needed and is stripped in icon code anyways ([BrightspaceUI/core/blob/main/components/icons/icon.js#L63](https://github.com/BrightspaceUI/core/blob/main/components/icons/icon.js#L63))


.nvmrc
```diff
@@ -0,0 +1 @@
+14
```
devpow112, line 1: Pinning to node 14/npm 6 as npm 8 has issues with installs and git style refs.

components/d2l-enrollment-hero-banner/d2l-enrollment-hero-banner.js
```diff
@@ -13,16 +13,16 @@ import 'd2l-organizations/components/d2l-organization-date/d2l-organization-date
 import 'd2l-organizations/components/d2l-organization-image/d2l-organization-image.js';
 import 'd2l-organizations/components/d2l-organization-info/d2l-organization-info.js';
 import 'd2l-typography/d2l-typography-shared-styles.js';
-import 'd2l-status-indicator/d2l-status-indicator.js';
+import '@brightspace-ui/core/components/status-indicator/status-indicator.js';
```
dlockhart, line 16: I was going to say this might fail linting due to the ordering... but everything is out of order so clearly linting isn't enabled here!

dlockhart (approved):
LGTM!

components/d2l-enrollment-hero-banner/d2l-enrollment-hero-banner.js
devpow112: Lol seems like. Maybe another task for another day (enable/update lint everywhere)

PR #253, opened 2022-02-10
sub old d2l-offscreen with core version

annabelsegalld2l:


components/d2l-enrollment-detail-card/d2l-enrollment-detail-card.js
```diff
@@ -279,7 +274,7 @@ class D2lEnrollmentDetailCard extends mixinBehaviors([
 			<d2l-resize-aware class="dedc-container" mobile$="[[_mobile]]">
 				<div class="dedc-base-container" has-link$="[[_organizationHomepageUrl]]">
 					<a class="d2l-focusable" href$="[[_organizationHomepageUrl]]">
-						<span class="dedc-link-text">[[_title]]</span>
+						<d2l-offscreen class="dedc-link-text">[[_title]]</d2l-offscreen>
```
dlockhart, line 277: Ditto here, I think the CSS class and associated CSS can go away too. `inline-block` on an offscreen element shouldn't be doing anything.

PR #256, opened 2022-02-11
Updated "Terminer" to "Terminé" for fr-ON

jeff-shepstone:


lloydwaddell (approved):
Neither of the translation PRs have a release prefix, can you merge this second and add one? You can do that by doing a squash and merge and adding "fix: " to the commit title

PR #265, opened 2022-03-08
remove unused Gulp and scripts

dlockhart:
So I believe at one point Gulp plus these other scripts were used to generate translation files in `./components/build/lang`... but I couldn't find anywhere that was actually calling these scripts, and the `build/lang` stuff has been modified to just import from `lang` directly.

Seems not needed.


package.json
```diff
@@ -7,8 +7,6 @@
   "name": "d2l-enrollments",
   "scripts": {
     "start": "es-dev-server --node-resolve --dedupe --watch --open /demo/",
-    "lang:update": "node scripts/updateLangJson.js",
-    "lang:build": "gulp build",
```
dlockhart, line 11: Nothing in CI runs these, and there's nothing in the README about running them. They also didn't work for me at all when I did run them. Seems safe to remove.

dlockhart:
This will fail CI for the same reason I'm switching this to not use `polymer-cli` -- so will need to force merge temporarily.

svanherk (approved):

dlockhart:
Possibly yeah... I'll have a look once I have things actually working again.

PR #269, opened 2022-03-09
US133580 - Adding OSLO support to d2l-enrollments repo

joshhoey:
Adding Oslo and Serge support to the `d2l-enrollments` component (which relies on many more components)

Language Package: *WebComponents*
Language Collection: **d2l-enrollments\Enrollments**

Note that the serge file is already in place for d2l-enrollments, but isn't mapped from the central bsi serge mapping file.


components/localize-behavior.js
```diff
@@ -0,0 +1,169 @@
...
+import da from './lang/da.js';
+import dadk from './lang/da-dk.js';
+import de from './lang/de.js';
+import en from './lang/en.js';
+import es from './lang/es.js';
+import eses from './lang/es-es.js';
+import fi from './lang/fi.js';
+import fr from './lang/fr.js';
+import frfr from './lang/fr-fr.js';
+import fron from './lang/fr-on.js';
```
dlockhart, line 15: Since you likely started this branch, there's now a Hindi file that should be included as well:
https://github.com/BrightspaceHypermediaComponents/enrollments/blob/master/components/lang/hi.js

components/localize-behavior.js
```diff
@@ -0,0 +1,186 @@
...
+				case 'en':
+					translations = en;
+					break;
+				case 'es': case 'es-es':
+					translations = es;
+					break;
+				case 'fi':
+					translations = fi;
+					break;
+				case 'fr': case 'fr-fr': case 'fr-ca':
```
bearfriend, line 139: This seems bad?

joshhoey: Yeah, I'm not wedded to this at all. I borrowed it from another file but I didn't like it (you and Stacey have already done good things in `d2-my-courses-ui`, so this can definitely follow that pattern)

dlockhart:
Closing in favour of #275.

PR #271, opened 2022-03-11
Cleanup unnecessary imports, try to add missing ones

svanherk:
While cleaning up the test logs, I tried to cleanup some style warning logs in the console as well.  I removed ones below that _should_ be safe, and added some missing ones that were only working because of other imported components happening to include that import.


components/d2l-enrollment-collection-widget/d2l-enrollment-collection-widget.js
```diff
@@ -63,7 +56,7 @@ class EnrollmentCollectionWidget extends EnrollmentsLocalize(EntityMixin(Polymer
 
 	static get template() {
 		return html`
-			<style include="d2l-typography-shared-styles">
```
svanherk, line 66: This isn't needed and causes the console warning.  I think we maybe used to the name the module?  But now we don't, and importing it seems to be enough based on my testing.  Does this match with what you'd expect?

components/d2l-enrollment-hero-banner/d2l-enrollment-hero-banner.js
```diff
@@ -1,13 +1,7 @@
 /**
 `d2l-enrollment-hero-banner`
-
-Polymer-based web component for a organization name.
-
-@demo demo/d2l-enrollment-hero-banner/d2l-enrollment-hero-banner-demo.html Organization Name
 */
 import { html, PolymerElement } from '@polymer/polymer/polymer-element.js';
-import 'd2l-link/d2l-link-behavior.js';
-import 'd2l-polymer-behaviors/d2l-focusable-behavior.js';
```
svanherk, line 10: From what I can tell these did nothing because we didn't hook up the behaviours...

components/d2l-enrollment-collection-widget/d2l-enrollment-collection-widget.js
dlockhart: So if it's a [Polymer 1 type "style module"](https://polymer-library.polymer-project.org/1.0/docs/devguide/styling#style-modules) that defines `<dom-module id="foo">` then I'd expect to see `<style include="foo">` when it's used. But the file they were importing doesn't define a `<dom-module>` at all, `d2l-typography.js` does (which in turn imports `d2l-typography-shared-styles.js`.

[Before it was blanked out](https://github.com/BrightspaceUI/typography/blob/96b1a7e89ef9b59aceac5e96f828c059045b6108/README.md), typography's README essentially said to import the main typography file and then just use the CSS classes, not the style mixins.

So I think this is the correct change.

components/d2l-enrollment-detail-card/d2l-enrollment-detail-card.js
```diff
@@ -61,7 +62,7 @@ class D2lEnrollmentDetailCard extends mixinBehaviors([
 
 	static get template() {
 		return html`
-			<style include="d2l-typography-shared-styles">
+			<style>
```
dlockhart, line 65: This is using `@apply --d2l-body-small-text;` down below... is that actually working?

svanherk: Yeah it does... if I remove that, these styles get removed:
```
    color: var(--d2l-body-small-text_-_color);
    font-size: var(--d2l-body-small-text_-_font-size);
    font-weight: var(--d2l-body-small-text_-_font-weight);
```
And if I comment out the `d2l-typography-shared-styles` import and the `d2l-sequences-module-list` import ([which also imports the `d2l-typography-shared-styles`](https://search.d2l.dev/xref/BrightspaceHypermediaComponents/sequences/components/d2l-sequences-module-list.js?r=df104956#11)) it stops adding the styles as well.

package.json
```diff
@@ -41,12 +41,12 @@
     "d2l-course-image": "github:Brightspace/course-image#semver:^3",
     "d2l-fetch": "github:Brightspace/d2l-fetch.git#semver:^2",
     "d2l-hypermedia-constants": "^6",
-    "d2l-link": "github:BrightspaceUI/link#semver:^5",
```
dlockhart, line 44: 🎉

dlockhart (approved):

package.json
```diff
@@ -41,12 +41,12 @@
...
     "d2l-localize-behavior": "github:BrightspaceUI/localize-behavior#semver:^2",
     "d2l-organization-hm-behavior": "github:Brightspace/organization-hm-behavior#semver:^3",
     "d2l-organizations": "github:BrightspaceHypermediaComponents/organizations#semver:^5",
     "d2l-resize-aware": "github:BrightspaceUI/resize-aware#semver:^1",
     "d2l-sequences": "github:BrightspaceHypermediaComponents/sequences#semver:^2",
+    "d2l-typography": "BrightspaceUI/typography#semver:^7",
```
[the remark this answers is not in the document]
svanherk: Good call, will update that PR when this merges

dlockhart (approved):

PR #273, opened 2022-03-15
fix: collection localization

dlockhart:
Enrollments has two Serge connections -- this only addresses the issues with the one, "enrollment-collection-view".

This switches it to use the `DynamicLocalizeMixin`, which means it'll be immune from any new language changes in the future. It also cleans up the translation files to match the default Serge settings -- they were copies of each other anyway.


components/d2l-enrollment-collection-view/lang/zh.js
```diff
@@ -1,13 +0,0 @@
-/* eslint quotes: 0 */
```
svanherk, line 1: Should this file be renamed to `zh-cn` instead of deleted?

svanherk: Lol nvm, that is now happening

dlockhart:
For some reason this isn't picking up the rename from `zh.js` -> `zh-cn.js`. Closing and recreating.

PR #275, opened 2022-03-15
feat: adding Oslo support to some enrollments elements

dlockhart:
Recreating #269 since there were too many conflicts after the linting fixes in #272.

I've also removed everything in the `build/lang` directory and cleaned up the translation files such that they match the default Serge config settings.

---

Adding Oslo and Serge support to the `d2l-enrollments` component (which relies on many more components).

Language Package: `WebComponents`
Language Collection: `d2l-enrollments\Enrollments`

Note that the serge file is already in place for `d2l-enrollments`, but isn't mapped from the central bsi serge mapping file.


components/d2l-enrollment-hero-banner/d2l-enrollment-hero-banner.js
```diff
@@ -22,7 +22,6 @@ import '@brightspace-ui/core/components/dropdown/dropdown-menu.js';
 import '@brightspace-ui/core/components/dropdown/dropdown-more.js';
 import '@brightspace-ui/core/components/menu/menu-item.js';
 import '@brightspace-ui/core/components/menu/menu-item-link.js';
-import { EnrollmentsLocalize } from '../EnrollmentsLocalize.js';
```
dlockhart, line 25: This wasn't needed since `DateTextAndStatusMixin` extends `EnrollmentsLocalize`.

test/d2l-enrollment-card.test.js
```diff
@@ -198,14 +198,6 @@ describe('d2l-enrollment-card', () => {
 			expect(component._canChangeCourseImage).to.equal(true);
 		});
 
-		it('should update _courseSettingsLabel', () => {
-			expect(component._courseSettingsLabel).to.equal('Course Name course settings');
-		});
-
-		it('should update _pinButtonLabel', () => {
-			expect(component._pinButtonLabel).to.equal('Course Name is pinned. Unpin course');
-		});
```
dlockhart, line 207: These tests aren't useful anymore now that these computed properties are gone.

bearfriend:
Did `lang/cy.js`and `lang/da.js` get lost in the mix?

bearfriend:
Ah, I see they already existed. I assume they're also up to date?

klhuff (approved):

components/build/lang/fi.js
```diff
@@ -1,15 +0,0 @@
-'use strict';
-
-import { dedupingMixin } from '@polymer/polymer/lib/utils/mixin.js';
-import { default as lang } from '../../lang/fi.js';
```
svanherk, line 4: We should be able to delete this file in `lang` as well.  It's all just English anyways, whereas My Courses actually had translations

svanherk:
> Did lang/cy.jsand lang/da.js get lost in the mix?
> Ah, I see they already existed. I assume they're also up to date?

@bearfriend Yeah if I'm reading the current setup correctly, serge updates the `cy.js` and `da.js` files (so they should be more up-to-date) but we _use_ the `cy-gb.js` and `da-dk.js`, which isn't great.  Doing a diff the changes are very small, so I think using the latest files that serge updated makes sense?

dlockhart:
> Did lang/cy.jsand lang/da.js get lost in the mix? Ah, I see they already existed. I assume they're also up to date?

Yeah they existed, and were the only ones actually hooked up to Serge -- in the history you can see that they've been updated recently vs. 9 months ago. I figured if Serge was controlling them, they'd be the "most correct".

components/build/lang/fi.js
dlockhart: Good catch -- removed.

bearfriend:
> > Did lang/cy.jsand lang/da.js get lost in the mix? Ah, I see they already existed. I assume they're also up to date?
> 
> Yeah they existed, and were the only ones actually hooked up to Serge -- in the history you can see that they've been updated recently vs. 9 months ago. I figured if Serge was controlling them, they'd be the "most correct".

Ah, right! Because of the new `output_lang_rewrite` defaults.

bearfriend (approved):

svanherk (approved):
🎉

